PR #30, opened 2019-10-15
[Rordev-54] changed: ssl external and internode documentation

pondzix:


elasticsearch.md
```diff
@@ -309,6 +311,50 @@ readonlyrest:
...
+In `readonlyrest.yml` following settings must be added (it's just example configuration presenting most important properties):
+
+```yml
+readonlyrest:
+    ssl_internode:
+      keystore_file: "keystore.jks"
+      keystore_pass: readonlyrest
+      key_pass: readonlyrest
+```
+Similar to `ssl` for HTTP, keystore should be stored in the same directory with `elasticsearch.yml` and `readonlyrest.yml`.
```
sscarduzio: "the" keystore

elasticsearch.md
```diff
@@ -309,6 +311,50 @@ readonlyrest:
...
+It can be done by means of:
+
+```yml
+certificate_verification: true
+```
+under `ssl_internode` section. This option is applicable only for internode ssl.
+   
+### Client authentication 
+
+By default client authentication is disabled. When enabled, server asks client about its certificate, so ES is able to verify client's identity. It can be enabled by means of:
```
sscarduzio: By default, THE client authenticaiton is disabled. When enabled, THE server asks THE client about [...] ..so ES is able to verify THE client's identity.

PR #41, opened 2020-01-16
[RORDEV-161]

coutoPL:
changes related to this PR: https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/546


elasticsearch.md
```diff
@@ -525,7 +525,13 @@ In ReadonlyREST we roughly classify requests as:
 * "write": when allowed, the request changes the internal state of the cluster or the data.
 
 If a **read request** asks for a some indices they have permissions for and some indices that they do NOT have permission for, the request is **rewritten** to involve only the subset of indices they have permission for.
-This is behaviour is very useful in Kibana: **different** users can see the **same** dashboards with data from only their own indices.|
+This is behaviour is very useful in Kibana: **different** users can see the **same** dashboards with data from only their own indices.
+
+When the subset of indices is empty, it means that user are not allowed to access requested indices. In multitenancy environment we should consider two options:
```
sscarduzio: * What subset? Define the set and the subset in question
* Remember to use the articles: i.e. "..that the users.."
* define better the user i.e. "the authenticated Kibana user" or something like this.

elasticsearch.md
```diff
@@ -525,7 +525,13 @@ In ReadonlyREST we roughly classify requests as:
...
+* requested indices don't exist
+* requested indices exist but belong to another user  
+
+For both these cases ROR is going to return HTTP 404 or empty response. The same behaviour will be observed for ES with ROR disabled (for nonexistent index). If some index does exist, but belongs to another user, ROR is going to pretend that the index doesn't exist and the response is the same like the index actually did not exist.
```
sscarduzio: * indices belonging to someone -> someone being authorized to see some indices
* say what is the response when the index actually does not esist.

elasticsearch.md
```diff
@@ -1233,6 +1239,15 @@ FORBIDDEN by default req={ ID:747832602--1038482600#1312150, TYP:SearchRequest,
...
 
+#### Requests finished with `INDEX NOT FOUND`
+
+This is an example of such request:
+
+```
+INDEX NOT FOUND req={  ID:501806845-1996085406#74,  TYP:GetIndexRequest,  CGR:N/A,  USR:dev1 (attempted),  BRS:true,  KDX:null,  ACT:indices:admin/get,  OA:172.20.0.1/32,  XFF:null,  DA:172.20.0.2/32,  IDX:nonexistent*,  MET:GET,  PTH:/nonexistent*/_alias/,  CNT:<N/A>,  HDR:Accept-Encoding=gzip,deflate, Authorization=<OMITTED>, Connection=Keep-Alive, Host=localhost:32773, User-Agent=Apache-HttpClient/4.5.2 (Java/1.8.0_162), content-length=0,  HIS:[CONTAINER ADMIN-> RULES:[auth_key->false]], [dev1 indexes-> RULES:[auth_key->true, indices->false], RESOLVED:[user=dev1]], [dev2 aliases-> RULES:[auth_key->false]], [dev3 - no indices rule-> RULES:[auth_key->false]]  }
+```
+
+The state above is only possible for read-only ES requests for blocks containing `indices` rule. If all other rules within the block are matched, but only `indices` rule is mismatched, the final state of the block is forbidden due to index not found.
```
sscarduzio: * define what a read-only request is for us (it's a ROR concept, not ES concept).
* articles

elasticsearch.md
coutoPL: the subset is defined in section above

coutoPL:
@sscarduzio wdyt about fixes?

sscarduzio:
Amazing, thank you @coutoPL! There is still some work to do above your changes (my old text) which is full of crappy typos. But I will take care of it after merging this. T

PR #43, opened 2020-02-08
changes: when prompt_for_basic_auth is set to true, ROR is going to r…

coutoPL:
…eturn 401 instead of 404 for all nonexistent indices or indices not allowed for given use

### DIFF
--- a/elasticsearch.md
+++ b/elasticsearch.md
@@ -544,6 +544,8 @@ When the subset of indices is empty, it means that user are not allowed to acces
 
 For both of these cases ROR is going to return HTTP 404 or HTTP 200 with an empty response. The same behaviour will be observed for ES with ROR disabled (for nonexistent index). If an index does exist, but a user is not authorized to access it, ROR is going to pretend that the index doesn't exist and a response will be the same like the index actually did not exist. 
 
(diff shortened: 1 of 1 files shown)


coutoPL:
@sscarduzio as 1.19.1 is already released, we should merge this PR. Do you have any comments? Do you want me to fix sth?

elasticsearch.md
```diff
@@ -544,6 +544,8 @@ When the subset of indices is empty, it means that user are not allowed to acces
 
 For both of these cases ROR is going to return HTTP 404 or HTTP 200 with an empty response. The same behaviour will be observed for ES with ROR disabled (for nonexistent index). If an index does exist, but a user is not authorized to access it, ROR is going to pretend that the index doesn't exist and a response will be the same like the index actually did not exist. 
 
+It's also worth to mentioned here, that when `prompt_for_basic_auth: true` (current default) is configured, ROR is going to return HTTP 401 error code instead of HTTP 404. It is relevant for ROR free version users who take advantage of Kibana's behaviour which shows the native browser basic auth dialog, when it receives HTTP 401 response.
```
sscarduzio: It's also worth mentionING... 
...that when  ``prompt_for_basic_auth` is set to `true` (that is, the default value)....

sscarduzio: ROR is going to return a 401 HTTP status code, instead of a 404 HTTP status code.

sscarduzio: Well, now that ROR Free for Kibana exists, this is not anymore so relevant :)

PR #66, opened 2020-09-30
Update kibana.md to show direct download links

askids:
Updated to show support for direct download links for specific version of plugin and  Elasticsearch.


sscarduzio:
thank you @askids

PR #73, opened 2020-12-03
[ROR 1.26.0][RORDEV-325] ror_internal_api rule doc

coutoPL:


elasticsearch.md
```diff
@@ -1111,6 +1111,34 @@ This authentication and authorization connector represents the secure channel \(
 
 Continue reading about this in the kibana plugin documentation, in the dedicated [SAML section](kibana.md#saml)
 
+### Protecting ROR internals
+
+#### `ror_internal_api`
+
+`ror_internal_api: "allow" # or "forbid"`
+
+ROR plugin has an internal API for an in-index settings management. The settings are stored in `.readonlyrest` index (or [custom one](./kibana.md#configuration)). Moreover, when an audit collector is enabled, audit entries will be stored in the audit index (by default in the index defined by a template `readonlyrest_audit-'yyyy-MM-dd`). All of these indices and the API should be protected and probably only an admin or a superuser should be able to have an access to it. To allow access to the ROR internal `ror_internal_api: "allow"` should be added to the admin/superuser's block. When `ror_internal_api` is not used in a given block, ROR implicitly adds `ror_internal_api: "forbid"` step which will be checked after block is matched.
```
sscarduzio, line 1120: the audit indices are not going to be inaccessible (for reads) by default , right? I mean, makes sense to make them not writeable directly by default. But if I want to allow read to them, I don't want to give implicitly the permission to delete/change ".readonlyrest".

kibana.md
```diff
@@ -831,6 +835,7 @@ readonlyrest:
     verbosity: error
     kibana_access: admin
     kibana_index: ".kibana"
+    ror_internal_api: allow
```
sscarduzio, line 838: Kibana does not need direct access to ".readonlyrest", nor it needs write access to "readonlyrest_audit-*". And we said custom logging API is accessible with regular users' credentials. So why this?
Actually, at this stage, why the name "ror_internal_api"? This is now about protecting RW access to .readonlyrest and W access to audit data. Neither of these activities are  mediated by any "internal ror api". :thinking:

elasticsearch.md
coutoPL: the new rule forbid an access when:
* it's NOT a readonly request to one of ROR indices (audit, in-index settings) OR
* it's a ROR config action (new API or old API)

so, yes - you don't have to add explicitly a permission to read eg. .readonlyrest index, but you cannot write to it without 
adding `ror_internal_api: "allow"` to the block.

kibana.md
coutoPL: I checked all our example configuration snippets and I added the new rule to each block which looks like an admin one. This one looks like it, doesn't it?

According to the name of the new rule. Yeap, I don't feel it's 100% hit, but this rule, when is going to check if:
* request is related to action `cluster:ror/config/manage` (new config API) or `cluster:ror/config/refreshsettings` (old config API) 
* request is about to change in-index index or audit index

So, I picked `ror_internal_api`. Maybe `ror_internals` is enough. As you can see we'd like to restrict access to internal config API and internal indices. 

Do you have a better name for the rule?

coutoPL: @sscarduzio wdyt?

sscarduzio: I think kibana_access: admin should be allowed to modify ROR settings via validated API, and at the same time **not** be allowed to read, or write ".readonlyrest". Direct access to this index should be reserved to apocalyptical situations where the index gets corrupted. Incidentally, direct access to that index is the cause of 90% of index corruption. 

I think the ror_inernal_api: allow should be called something like `access:  [ maintenance ]` (thus, avoiding the creation of a rule with boolean value, which is inherently limiting for the future). 

`access` rule can have value:
  - `maintenance`: can do anything
  - `ror_admin`: no direct access (R or W) to ".readonlyrest"  (THIS IS THE IMPLICIT DEFAULT FOR ANY BLOCK)
  - `ror_user`:  no direct access (R or W) to ".readonlyrest", forbidden "cluster:ror/*" actions (except custom audit logs creation
 
The "kibana_access" rule will still stand for now, as it's another thing entirely and includes other limitations. Combining access and "kibana_access" should probably raise a warning though, as most likely the bigger strictness of any kibana_access rule will overpower the "access" rule in the economy of any ACL block.

### Things a typical existing ROR user will need to adjust after this PR
We should obviously strive to minimize the amount of these. 

* Add ` `access: maintenance` to any failsafe user lying around in readonlyrest.yml in case they need to delete ".readonlyrest" index when they forget their admin password.

WDYT?

coutoPL: the solution was redesigned in RORDEV-401. I'm closing this one because it's not needed any more

PR #81, opened 2021-01-13
Feat/oidc additional claims and groups

pchesneau:
Document how to access custom claims in OIDC userinfo token + examples to use groupsParameter


sscarduzio:
Thank you @pchesneau!

PR #82, opened 2021-01-15
[RORDEV-356] filter fields not matching ror requests

pielas:
I don't know if such description is enough. I will think in the meantime how can I improve it. If you have some ideas please let me know

### DIFF
--- a/elasticsearch.md
+++ b/elasticsearch.md
@@ -785,6 +785,8 @@ In this example, we want to avoid that users belonging to group "press" can see 
 
 **⚠️IMPORTANT** The `filter`and `fields` rules will only affect "read" requests, therefore "write" requests **will not match** because otherwise it would implicitly allow clients to "write" without the filtering restriction. For reference, this behaviour is identical to x-pack and search guard.
 
+**⚠️IMPORTANT** Beginning with version 1.27.0 all ROR internal requests from kibana will not match blocks containing `filter` and/or `fields` rules.
+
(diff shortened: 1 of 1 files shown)


elasticsearch.md
```diff
@@ -779,6 +779,8 @@ In this example, we want to avoid that users belonging to group "press" can see
 
 **⚠️IMPORTANT** The `filter`and `fields` rules will only affect "read" requests, therefore "write" requests **will not match** because otherwise it would implicitly allow clients to "write" without the filtering restriction. For reference, this behaviour is identical to x-pack and search guard.
 
+**⚠️IMPORTANT** Beginning with version 1.27.0 all ROR internal requests from kibana will not match blocks containing `filter` and/or `fields` rules.
```
coutoPL, line 782: maybe we should clarify that ROR internal API is used by ROR Kibana during login & config reloading?

coutoPL (changes requested):

elasticsearch.md
sscarduzio: I think it's worth to show the example from tests you showed us in the Jira ticket. It very much gives the idea of the ramifications this change has.

pielas: I added some additional info and changed examples of "block duplication strategy"

coutoPL (approved):
For me it's OK. @sscarduzio please let us know if it's also ok for you or maybe there is sth you'd like to add?

sscarduzio:
LGTM, thanks

PR #85, opened 2021-02-16
username case mapping

clutroth:


elasticsearch.md
```diff
@@ -1520,6 +1510,17 @@ The `groups` rule accepts a list of group names. This rule will match if the res
 
 _Example: rules are associated to groups \(instead of users\) and users-group association is declared separately later under `users:`_
 
+### Username case sensitivity
+ReadonlyREST can cooperate with services, that operates in case-insensitive way.
+For this case ror has toggleable username case sensitivity option `username_case_sensitivity`.
```
coutoPL, line 1515: ROR

elasticsearch.md
```diff
@@ -1520,6 +1510,17 @@ The `groups` rule accepts a list of group names. This rule will match if the res
...
+
+```text
+readonlyrest:
+  username_case_sensitivity: case_sensitive
+ ```  
+By default, usernames are case-sensitive `username_case_sensitivity: case_sensitive`.
+By setting `username_case_sensitivity: case_sensitive` every username comparison will be case-insensitive.
```
coutoPL, line 1522: maybe instead of "every" we should say "in any rule"?

coutoPL (changes requested):

PR #88, opened 2021-03-17
[RORDEV-454] detailed description of templates requests handling process of `indices` rule

coutoPL:


elasticsearch-details/indices-rule-templates.md
```diff
@@ -0,0 +1,74 @@
+# Templates handling in `indices` rule
```
sscarduzio, line 1: This part is very analytical, like code translated in English. Despite it being precise, it lacks examples entirely and is not goal oriented.
This is in stark contrast with the "index not found" document above, which reads way smoother.

elasticsearch-details/indices-rule-templates.md
```diff
@@ -0,0 +1,74 @@
...
+  
+### Delete a component template
+
+The request will be allowed when template does not exist OR all of the following conditions are met:
+* a template with requested name does exist,
+* all aliases of the existing template are allowed.
+  
+### Get component templates
+
+At the moment there is no way to restrict which component templates can be visible to the user (it is going to change in the
```
sscarduzio, line 67: yeah, show me!

elasticsearch.md
```diff
@@ -528,141 +528,20 @@ When the subset of indices is empty, it means that user are not allowed to acces
...
->   HTTP/1.1 200 OK
->   content-type: application/json; charset=UTF-8
->   content-length: 4
->
->   { }
-> ```
->
-> As we see, response is the same as we have experienced when there was really no index matching the pattern. Also here a user has a feeling that only his indices are present on a cluster.
+Templates are also connected with indices, but rather indirectly. 
+Index template has index patterns and could also have aliases.
```
sscarduzio, line 540: Either: Index templates have index patterns...
OR: An index template has index patterns...

sscarduzio (commented):
Precise, but needs a change of perspective: try writing from the user side. Fpr example you can have a sort of dialogue between an explorative voice: "let's try this... Why did it behave like this?" and a omniscient voice "that didnt work because ROR works like this" that keep the writing purposeful, reliable and informative. Almost narrative.

elasticsearch-details/indices-rule-templates.md
coutoPL: @sscarduzio  I've added example in one section. Is it ok and should I continue to add rest of examples in this style?

elasticsearch-details/indices-rule-templates.md
```diff
@@ -12,6 +36,93 @@ The request will be allowed when all of following conditions are met:
...
+
+  {
+    "acknowledged" : true
+  }
+```
+
+Hooray! The index template was added. This time ROR allowed us to do so. It's because `dev1` user has an access to index pattern `idev1_test*`, because it is contained in `idev1_*`. Used aliases are also allowed. 
+
+</details>
+
```
sscarduzio, line 125: Yeah, this is really nice :)

sscarduzio (commented):
Yes, go ahead with examples like that :+1:

sscarduzio (approved):
I really like this style, it's very clear and reading flows better.

sscarduzio:
shouldn't this be merged against develop?

coutoPL:
wow, not finished yet. I need to add several examples of component templates. There are several differences which should be mentioned. 

And yes, I will merge changes to develop and master too.

PR #100, opened 2021-11-09
[RORDEV-552] Impersonation docs

coutoPL:
From the perspective of ROR ES only, impersonation is not supported by many rules. Until now there was no word about impersonation in our docs. Now, I've added a separate file with the description of the feature. 

ROR impersonation shines but only with ROR Auth Mock API and Test Settings API. The APIs are ROR internal ones and will be used by Kibana in the future. Also in the future, we will describe, from the perspective of Kibana, how to achieve full impersonation support.

### DIFF
--- a/elasticsearch-details/impersonation.md
+++ b/elasticsearch-details/impersonation.md
@@ -1,0 +1,107 @@
+# Impersonation 
+
+According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
+
+> An impersonator is someone who imitates or copies the behavior or actions of another.
+
+So, an impersonation can be understood as imitating behaviors or actions.
+In the context of ReadonlyREST eg. one user could imitate an action 
+of another user. Why would we want it? Let's suppose the first user is 
+an admin, who has just configured access for a new user. He would like 
+to know if the rule or rules are configured correctly. And here it comes the impersonation feature. The admin can call ES request on behalf of the 
+tested user and authenticate using his credentials. 
+
+## Impersonators configuration
+
+Not anyone can be the impersonator. The list of impersonators and their permissions are defined in `impersonation` section:
+
+```yaml
+readonlyrest:
+  
+  access_control_rules:
+
+    [...]
+
+  impersonation:
+
+    - impersonator: admin1      // impersonator user name or pattern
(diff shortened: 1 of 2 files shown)


elasticsearch-details/impersonation.md
```diff
@@ -0,0 +1,107 @@
...
+>    ],
+>    "type":"forbidden_response",
+>    "reason":"forbidden",
+>    "due_to":["OPERATION_NOT_ALLOWED", "IMPERSONATION_NOT_ALLOWED"]
+>  },
+>  "status":403
+>}
+>```
+
+Some authentication or/and authorization rules (like LDAP ones or related to external systems) don't support impersonation by default (the support will be possible with cooperation with Kibana plugin soon). When request with impersonation
```
sscarduzio, line 84: I don't understand what "by default" means here. And I don't understand how the support for external auth rules will be enabled by the Kibana plugin in the future.

coutoPL: You know, impersonation can be easily supported by the rules that have static users and/or groups. If a rule depends on an external system, like an LDAP, you have two ways:
1) depend on impersonation supported by the system
2) mock response from the system

We picked the 2nd way. There is ROR Auth Mock API that can be used to configure the mocked response from LDAP. Eg. ldap1 has users u1 and u2 with groups: [g1, g2] and [g1, g3]. The API at this moment should be considered as an internal one - that's why it's described only in ROR internal docs. Kibana could take advantage of the API (the flow was agreed with the customer) and configure mock responses for the LDAP connector for the tests with impesonation. 

To sum up:
eg. `ldap_auth` rule doesn't support impersonation by default, but then you configure, using the internal Auth Mock API, the LDAP connector mocked users and groups, the impersonation will be supported by the rule.

sscarduzio: OK So because the configuration is supported via mock API only, and because mock API is internal, we tell the user wait for our Kibana UI if you want to impersonate. 

Do we have a rough idea of how this UI should visually look like? Is there any  similar UI in any other software we can get familiar with and  take as an example?

coutoPL: there is a Jira for the topic. Let's move the discussion there. RORDEV-484

PR #103, opened 2021-12-02
Added some info about PKCS#12 certificate support

pielas:


sscarduzio:
Thanks, I took the opportunity to rework other parts of this section a bit.

PR #104, opened 2021-12-20
Update groups-rule-mapping.md

sscarduzio:
#TODO @coutoPL what's the definition/difference between **common** and **detailed** group mapping? 🧐

### DIFF
--- a/elasticsearch-details/groups-rule-mapping.md
+++ b/elasticsearch-details/groups-rule-mapping.md
@@ -1,15 +1,15 @@
 # External to local groups mapping 
 
-The `groups` accepts a list of group names. This rule will match requests in which the resolved username belongs at least to one of the listed groups. The association between usernames and groups is explicitly declared in the users section of the ACL. This is a list of usernames (full wildcard patterns are supported).
+The `groups` rule accepts a list of group names. This rule will match on a requests in which the resolved username belongs at least to one of the listed groups. The association between usernames and groups is explicitly declared in the users section of the ACL. This is a list of usernames, and today, full wildcard patterns are supported.
 
-Each entry in the association rule requires:
-* an authentication rule: (I.e. auth_key_* for local credentials, or ldap_authentication, external_authentication, etc)
+Each entry in the `groups` rule (in the users section) requires:
+* an authentication rule: (I.e. one of the `auth_key_*` rules for local credentials, or `ldap_authentication`, `external_authentication`, etc)
 * a list of groups within the ones precendently inserted in the groups rules of the ACL blocks
-* optional: a remote authorization rule (ldap_authorization, groups_provider_authorization, etc.)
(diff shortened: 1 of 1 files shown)


coutoPL:
> #TODO @coutoPL what's the definition/difference between **common** and **detailed** group mapping? 🧐

there is no definition provided but the difference is shown in the example: https://github.com/beshu-tech/readonlyrest-docs/blob/master/elasticsearch-details/groups-rule-mapping.md#example

elasticsearch-details/groups-rule-mapping.md
```diff
@@ -1,15 +1,15 @@
...
+The `groups` rule accepts a list of group names. This rule will match on a requests in which the resolved username belongs at least to one of the listed groups. The association between usernames and groups is explicitly declared in the users section of the ACL. This is a list of usernames, and today, full wildcard patterns are supported.
 
-Each entry in the association rule requires:
-* an authentication rule: (I.e. auth_key_* for local credentials, or ldap_authentication, external_authentication, etc)
+Each entry in the `groups` rule (in the users section) requires:
+* an authentication rule: (I.e. one of the `auth_key_*` rules for local credentials, or `ldap_authentication`, `external_authentication`, etc)
 * a list of groups within the ones precendently inserted in the groups rules of the ACL blocks
-* optional: a remote authorization rule (ldap_authorization, groups_provider_authorization, etc.)
+* optional: a remote authorization rule (`ldap_authorization`, `groups_provider_authorization`, etc.)
```
coutoPL, line 8: remote? can `jwt_auth` be called `a remote rule`?

sscarduzio: good point!

coutoPL: why not just the `authorization rule`? Any authorization rule is supported. 

And one more thing. Take a look at [the example](https://github.com/beshu-tech/readonlyrest-docs/blob/cf60712d20cbf66c47e3b0dfc360253801105e74/elasticsearch-details/groups-rule-mapping.md#example). 

In ROR we have:
* authentication rules (can authenticate users)
* authorization rules (can authorize users)
* auth rule (can authenticate and authorize user)

In general users section looks like this:
```
users:
  - username: "..."
    groups: [...]
    <any_authentication_rule>
    
  - username: "..."
    groups: [...]
    <any_authentication_rule>
    <any_authorization_rule>
    
  - username: "..."
    groups: [...]
    <any_auth_rule> # excluding `groups` which is one of the auth rules, but we don't want it to be used here
```

So, according to the example and the general schema of the rule (described also [here](https://github.com/beshu-tech/readonlyrest-docs/blob/cf60712d20cbf66c47e3b0dfc360253801105e74/elasticsearch.md#groups)), the description is not complete

PR #105, opened 2021-12-26
[RORDEV-431] Custom audit cluster docs

pondzix:


coutoPL (approved):
LGTM @sscarduzio do you want to add any comments?

elasticsearch.md
```diff
@@ -1469,6 +1475,20 @@ logger.access_log_rolling.filter.regex.onMatch = DENY
...
 
+### Custom audit cluster
+
+It's possible to set a custom audit cluster responsible for audit logs storage. When custom cluster is specified, items will be sent to defined cluster nodes instead of the local one.
+
+```text
+readonlyrest:
+  audit:
+    collector: true
+    cluster: ["https://auditNode1:9200", "https://auditNode2:9200"]
```
sscarduzio, line 1486: How to use credentials when connecting to  those remote nodes? What kind of authentication is supported?

coutoPL: there is no authentication ATM. We can add basic auth (eg. `https://admin:pass@auditNode1:9200`) to the URL but in the future (when we will use ROR-Authorization header instead of Authorization one) we will be forced to extend the cluster address definition to pass custom header). Is it ok?

sscarduzio: I think showing this configuration example without any authentication in our documentation is not admissible. The basic Auth credentials within the URL is ok as syntax. But we need to understand that being able to configure the http client, e.g. things like custom headers, SSL cert, keystore, etc are a very common requirements nowadays and we should support them on any part of our plugin that resembles an http client. Possibly reusing the configuration schemas, so we harmonise the way any http client is configured in ror

coutoPL: yeah, I wanted to mention that we don't have this at the moment, but it looks like our HTTP clients' configurations should be richer. I've created a task for it. 

According to the audit cluster - basic auth support was added to docs and code.

sscarduzio (approved):

PR #109, opened 2022-03-01
Update elasticsearch.md

sscarduzio:
add documentation


groups-rule-mapping.md
```diff
@@ -0,0 +1,118 @@
+# External to local groups mapping
```
coutoPL, line 1: this file is present in `elasticsearch-details` folder

coutoPL (changes requested):
I have only one comment. We can merge when it's fixed

groups-rule-mapping.md
sscarduzio: Oh shoot! Missed this.

PR #110, opened 2022-03-12
[RORDEV-619] patching/unpatching ES > 8.0.x

coutoPL:


elasticsearch.md
```diff
@@ -64,7 +64,22 @@ Notice how we need to type in the format `file://` + absolute path \(yes, with t
 
 When prompted about additional permissions, answer **y**.
 
-#### 3.Create settings file
+#### 3. Patch ES
+
+If you are using Elasticsearch 8.0.x or newer, you need **an extra post-installation step**. This will add some X-Pack jars to ROR's installation directory.
```
sscarduzio, line 69: s/add/copy

elasticsearch.md
```diff
@@ -161,14 +191,27 @@ e.g.
 bin/elasticsearch-plugin install file:///tmp/readonlyrest-1.16.15_es6.1.1.zip
 ```
 
-#### 4. Restart Elasticsearch.
+#### 5. Patch ES
+
+If you are using Elasticsearch 8.0.x or newer, you need **an extra post-installation step**. This will add some X-Pack jars to ROR's installation directory.
```
sscarduzio, line 196: s/add/copy

elasticsearch.md
```diff
@@ -193,13 +236,28 @@ service stop elasticsearch
 
 depending on your environment.
 
-#### 2. Uninstall ReadonlyREST from Elasticsearch:
+#### 2. Unpatch ES
+
+If you are using Elasticsearch 8.0.x or newer, you need **an extra post-installation step**. This will add some X-Pack jars to ROR's installation directory.
```
sscarduzio, line 241: Better say: "this will bring the Elasticsearch installation back to its original state"

As today it only does this copy/delete jar thing. In the future we'd like to do all crazy shit with patch tool.

elasticsearch.md
```diff
@@ -143,13 +158,28 @@ service stop elasticsearch
 
 depending on your environment.
 
-#### 2. Uninstall ReadonlyREST
+#### 2. Unpatch ES
+
+If you are using Elasticsearch 8.0.x or newer, you need **an extra pre-uninstallation step**. This will remove all additional files from ROR's installation directory.
```
sscarduzio, line 163: Better say: "this will bring the Elasticsearch installation back to its original state"

As today it only does this copy/delete jar thing. In the future we'd like to do all crazy shit with patch tool.

sscarduzio (changes requested):

elasticsearch.md
coutoPL: I had a problem with that, because:
1. Potentially for different ES versions patching can be differently implemented, so maybe it's ok to say, that we need to modify ES/ROR a little bit, but without telling what we do exactly
2. But because ROR is a security plugin, maybe the user would like to know what exactly we do during the patching

Nevertheless, in the "Patch ES" section we're telling the user about "copying the jars", so here I change it to "removing  them". Is it ok?

elasticsearch.md
coutoPL: same as above

elasticsearch.md
sscarduzio: Oh sure, yeah we should change the Patch ES section! We should say that: "depending on the Elasticsearch version, this command might tweak the main Elasticsearch installation files and/or copy some jars to `plugins/readonlyrest directory". A link to the current patcher class implementation can be added, let's make sure we have plenty of comments there.

sscarduzio (approved):

PR #114, opened 2022-05-17
Rordev 671 add impersonation feature description

Dzuming:
- Provided impersonation feature description


kibana/README.md
```diff
@@ -996,3 +996,64 @@ readonlyrest:
...
+
+To set test settings:
+
+1. Open the RoR menu
+2. Click the Edit security settings button
+
+![Test settings ror menu](<../.gitbook/assets/test_settings_ror_menu.png>)
+
+3. Go into the Test settings tab
+4. You can set the TTL value. After the TTL value is reached, your test settings will be deactivated automatically
```
sscarduzio, line 1021: Acronyms should be expanded and described.

"You can optionally (🙋‍♂️ is it optional? If yes, mention the default behaviour) set the "time to live" (TTL), that is a time interval after which the test settings will be automatically deactivated (🙋‍♂️ will the impersonation session also abruptly exit?)".

kibana/README.md
```diff
@@ -996,3 +996,64 @@ readonlyrest:
...
+
+#### Local and auth mock services users configuration
+
+1. Open the RoR menu
+2. Click the Edit security settings button
+
+![Impersonate ror menu](<../.gitbook/assets/test_settings_ror_menu.png>)
+
+3. Go into the Impersonate tab
+4. You can free type impersonate. This button is available only in the situation when in some cases, the system is not able to receive all usernames. In this case, to impersonate, you need to type impersonating username manually.
```
sscarduzio, line 1039: "You can freely type a username to impersonate using the dedicated button. This button will be available only when the system cannot anticipate all the possible usernames (for example, when ROR is configure to infer usernames from headers, etc)." 

Correct this if wrong, but the idea is to expand the explanation providing some scenarios.

kibana/README.md
```diff
@@ -996,3 +996,64 @@ readonlyrest:
...
+#### Add/edit auth mock user
+
+After clicking add/edit user buttons(5), you will see a dialog with an option to add(6) or remove(7) user from external auth mock
+
+![Add/edit auth mock service](<../.gitbook/assets/add_edit_auth_mock_service.png>)
+
+
+#### Impersonate a user
+
+1. After successfully impersonating, you should see impersonating username
```
sscarduzio, line 1054: When an impersonation session is started correctly, the "impersonating" will be visible in the RoR menu a shown in the picture.

sscarduzio (commented):
some suggestions!

kibana/README.md
```diff
@@ -996,3 +996,64 @@ readonlyrest:
 ```
 
 Now try to login as user1, and ReadonlyREST Enterprise should initialise the index ".kibana\_user1" with all the index patterns and dashboards contained in the template tenancy.
+
+## Impersonation
```
coutoPL, line 1000: shouldn't we have a separate page for the Impersonation feature? Obviously, at this level, we could add info that there is sth like Impersonation, what is this and what scenarios can it be useful for 

(we can extend the first paragraph from https://github.com/beshu-tech/readonlyrest-docs/blob/de65586040c8c20c2f77fa725bdf567d60d6f969/elasticsearch-details/impersonation.md)

kibana/README.md
```diff
@@ -996,3 +996,64 @@ readonlyrest:
...
+
+User with specific roles (`admin` or `unrestricted`) and permissions (`impersonation` section in the `readonlyrest.yml` file)
+can impersonate (imitate behaviors or actions of the user) another user.
+
+### Test settings
+
+During user impersonation, the `reasonlyrest.yml` settings are replaced by test settings and used during a whole impersonation process
```
coutoPL, line 1007: I don't think we should use a statement a `readonlyrest.yml` settings. It suggests that the settings are placed in the file, but it also works for in-index settings. 

I'd recommend using `ROR settings` instead.

kibana/README.md
```diff
@@ -996,3 +996,64 @@ readonlyrest:
...
+You need to configure test settings to be able to impersonate the user
+
+
+#### Setting test settings
```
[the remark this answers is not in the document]
coutoPL: shouldn't we explain why?

This is important that the main (production) ROR engine is unaffected. It means that a user/admin is able to modify the test settings without worrying that their actions would break sth. 

But I'm not sure we should use vocabulary like "ROR engine". Any ideas on how to call it?

kibana/README.md
```diff
@@ -996,3 +996,64 @@ readonlyrest:
 ```
 
 Now try to login as user1, and ReadonlyREST Enterprise should initialise the index ".kibana\_user1" with all the index patterns and dashboards contained in the template tenancy.
+
+## Impersonation
+
+User with specific roles (`admin` or `unrestricted`) and permissions (`impersonation` section in the `readonlyrest.yml` file)
+can impersonate (imitate behaviors or actions of the user) another user.
+
+### Test settings
```
coutoPL, line 1005: For me, it'd be nice to have a section where we describe what the Impersonation feature looks like.  I mean, that we could tell a user/admin how to configure sth (like test settings), but first they should be aware of what is this. And why they even have to do it.

kibana/README.md
```diff
@@ -996,3 +996,64 @@ readonlyrest:
...
+
+During user impersonation, the `reasonlyrest.yml` settings are replaced by test settings and used during a whole impersonation process
+You need to configure test settings to be able to impersonate the user
+
+
+#### Setting test settings
+
+To set test settings:
+
+1. Open the RoR menu
```
coutoPL, line 1015: ROR

kibana/README.md
coutoPL: the same here - maybe we should explain somewhere why the TTL was introduced.

kibana/README.md
```diff
@@ -996,3 +996,64 @@ readonlyrest:
...
+3. Go into the Test settings tab
+4. You can set the TTL value. After the TTL value is reached, your test settings will be deactivated automatically
+5. You can load current settings as test settings
+6. You can deactivate settings manually
+7. You can save test settings as settings
+
+![test settings tab](<../.gitbook/assets/test_settings_tab.png>)
+
+### Impersonate
+It contains a list of external services and local users. You can add and edit the list of users in external services and impersonate a specific user.
```
coutoPL, line 1029: we should explain somewhere why a user/admin needs to define external services' users and groups. 

Maybe we should use the LDAP service as an example. 
Maybe @mateuszkp96 should help to describe it.

kibana/README.md
```diff
@@ -996,3 +996,64 @@ readonlyrest:
...
+
+
+#### Impersonate a user
+
+1. After successfully impersonating, you should see impersonating username
+2. Click the Finish impersonation button to stop impersonation and go back into Impersonate tab
+
+All logs of impersonated user will have this format `[<log level>][plugins][ReadonlyREST][<filename>][impersonating <impersonated user username>]`
+
+![Impersonate user](<../.gitbook/assets/impersonate_impersonated_menu.png>)
```
coutoPL, line 1059: It looks like we also need a section where we mention about limitations of the current implementation of the feature:
1. rules that don't support it
2. the fact that `test settings` are stored in the memory of the node that handled the PUT tests settings request
3. the missing ("unknown") users that have to be entered manually 
4. ???

kibana/README.md
```diff
@@ -996,3 +996,64 @@ readonlyrest:
 ```
 
 Now try to login as user1, and ReadonlyREST Enterprise should initialise the index ".kibana\_user1" with all the index patterns and dashboards contained in the template tenancy.
+
```
coutoPL, line 999: @mateuszkp96 to describe the configuration of ROR settings related to the Impersonation feature you can use [this](https://github.com/beshu-tech/readonlyrest-docs/blob/de65586040c8c20c2f77fa725bdf567d60d6f969/elasticsearch-details/impersonation.md). We don't have to mention any more about the requests format (now we treat it as our internal API)

coutoPL (changes requested):

sscarduzio:
On a side note: should we consider making this a separate dedicated page? Load times of the main ES docs page are stellar already.

kibana/README.md
Dzuming: If I understand correctly, you suggest:

1. Add some basic info about the impersonation feature into the  ReadonlyREST docs kibana section
2. Move the rest into the markdown file on the GitHub 

I understand the motivation, the downside of this approach is the reduction of the ability to search for a specific phrase related to impersonation, like (auth mock service, test settings, free type impersonation, etc.), since the "advance" description will be in the separate GitHub file. Maybe it's a wrong assumption but, it will be hard  for the user to discover, that there is another file with a feature description

I'm not sure if you are talking about the same when you mention "separate page" (probably yes, but better to ask :P ) https://github.com/beshu-tech/readonlyrest-docs/pull/114#issuecomment-1129096503

kibana/README.md
Dzuming: done ✔️

kibana/README.md
Dzuming: To me, looks good, one question to @coutoPL, is it the correct example :) ? `for example, when ROR is configure to infer usernames from headers, etc`

kibana/README.md
sscarduzio: Yes, a small paragraph in main page and a whole page with the details. The search function is global in gitbook anyway.

kibana/README.md
coutoPL: You, as an admin, can use variables in ROR rules. There is a variable that allows you to extract value from headers. It can be used in context of some auth rule. In this case we cannot return users from the rule because their names are dynamic. 

So, yes - this is a correct example

kibana/README.md
Dzuming: `is it optional? If yes, mention the default behavior` It was optional, but based on recent changes, it's required now,

image


`will the impersonation session also abruptly exit?`

That's a very good question, Looks like we need to fix it because in the current approach we log out a user (btw we need to fix it because looks like, for some reason, /session-probe endpoint return 200 status, and we probably don't send `x-ror-impersonating` header in this case)

kibana/README.md
coutoPL: we should take into consideration the fact that the user will read [this](https://docs.readonlyrest.com/), not the GitHub repo markdowns directly

kibana/README.md
coutoPL: I hope that @mateuszkp96 will help with these

kibana/README.md
Dzuming: Ok, so this [Github markdown page](https://github.com/beshu-tech/readonlyrest-docs/blob/de65586040c8c20c2f77fa725bdf567d60d6f969/elasticsearch-details/impersonation.md) will be the part of the [gitbook](https://docs.readonlyrest.com/).

 Because if I understand correctly, now:
1. Impersonation section is not a part of [gitbook](https://docs.readonlyrest.com/)
2. The only reference to this impersonation description  file is a hyperlink, for example here https://docs.readonlyrest.com/elasticsearch#jwt_auth

sscarduzio: Correct, let's change that so we have 1/2sentences about impersonation (enough so people ctrl f), and then a full blown page a out it separately

kibana/README.md
sscarduzio: I don't remember how you implemented this, and I'm from my phone now. But session probe calls to pkp, then pkp goes to es to refresh the session iirc. So the impersonation header is ok if it's handled by the esclient as part of the authentication headers stored in the session. Right?

Dzuming: I changed this logic recently, so

1. /session-probe calls to pkp
2. If "impersonation" mode we ask elasticsearch if the test settings ttl is still valid, and if not, we finish the impersonation session and redirect a user into the impersonation tab

All changed session-probe logic is here https://github.com/sscarduzio/readonlyrest_kbn/pull/303/files#diff-224f72aa407e80ab0821d8da5391a1a0b8ec114175355a10f11048abd749363dR163

kibana/README.md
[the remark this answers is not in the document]
Dzuming: Ok, What about:

`For impersonation to work, some valid test settings should be created and saved. It's important that the main ROR setting will be unaffected, so you, as an admin/user don't need to worry that you will break something. Here is how to write  test settings:`
?

coutoPL:
https://github.com/sscarduzio/readonlyrest_kbn/pull/303#pullrequestreview-982231327

Great :) IMO you @here-s-johnny (and maybe @pielas) will be a great reviewer of the doc when we agree a final version from our perspective 👍

kibana/README.md
Dzuming: text changed ✔️

kibana/README.md
Dzuming: text changed ✔️

kibana/README.md
Dzuming: changed ✔️

kibana/README.md
Dzuming: > the same here - maybe we should explain somewhere why the TTL was introduced.

@coutoPL And what is the reason to introduce TTL? :P If I understand correctly the reason to introduce it is an optimization of resources, if several users will have test engines loaded, it will block resources somehow (?)

kibana/README.md
Dzuming: @mateuszkp96 did you have a chance to take a look into it :)?

kibana/README.md
Dzuming: @mateuszkp96 did you have a chance to take a look into it :)?

kibana/README.md
coutoPL: The resources are not so important, but it's true. We just want the admin to invalidate the Test Settings after his work. It's just a helper feature for him (aka playground where he is allowed to experiment without breaking the production cluster). For security reasons we also don't want to have the playground active for a long time (because he could forget about the invalidation of the Test Settings)

Dzuming: Ok, so the value for the admin is a security, so maybe we can add something like:

 ` "Time to live" (TTL) is introduced to automatically deactivate test settings when your work is finished. For security reasons, it's always good to deactivate your testing playground, and "time to live" (TTL) will ease this process. `

Maybe it's good to write what exactly the security reasons are? (I'm not sure)

Dzuming:
In this pr I I also reduced the size of all assets, thanks to this tool https://tinypng.com

from
image


to
image

So we are able to save some space, without losing the quality of images

kibana/README.md
Dzuming: Moved all content into impersonation.md file and left a few sentences and links in a gitbook

sscarduzio:
Just yesterday I removed hundreds of GB of trial builds from S3, so this change is very welcome.

elasticsearch-details/impersonation.md
```diff
@@ -50,14 +50,14 @@ curl -vk -u user1:user_secret "https://localhost:9200/twitter/_search?pretty"
...
+curl -vk -u admin1:pass -H "x-ror-impersonating: user1" "https://localhost:9200/twitter/_search?pretty"
 ```
 
-As we can see the request looks pretty much the same. The main difference is the `impersonate_as` header that contains `user1`, and authorization header value has the admin's credentials now. 
+As we can see the request looks pretty much the same. The main difference is the `x-ror-impersonating` header that contains `user1`, and authorization header value has the admin's credentials now. 
 
 Responses of both requests should be the same.
 
-Should the admin pass wrong credentials, or should they not be alowed to impersonate the user provided in the `impersonate_as` header, 
+Should the admin pass wrong credentials, or should they not be alowed to impersonate the user provided in the `x-ror-impersonating` header,
```
sscarduzio, line 60: allowed

elasticsearch-details/impersonation.md
```diff
@@ -138,6 +138,9 @@ For impersonation to work, some valid test settings should be created and saved.
 ![Impersonate tab](<../.gitbook/assets/impersonate_tab.png>)
 
 ## Add/edit auth mock user
+External services mock is used to simulate the response of existing authentication or authorization service like LDAP.
+You don't need to create a user account for configuration testing.
+You will only need to define users (and the groups which belong) returned by external services, listed in test settings.
```
sscarduzio, line 143: ...(and their associated groups) that would normally be returned by the external services...

elasticsearch-details/impersonation.md
```diff
@@ -151,4 +154,18 @@ After clicking add/edit user buttons(5), you will see a dialog with an option to
...
+
+* Not all features available in the ROR configuration are testable with impersonation mode.
+Some rules used in ROR ACL do not support impersonation. For example, auth rule with hashed credentials(e.g. `auth_key_sha512`) can be used in impersonation mode only when credentials follow the format `USER_NAME: HASH(PASSWORD)`; A fully hashed username and password don't allow fetching a username. The auth rule in such a format won't match during impersonation.
+A list of the rules with info about impersonation support can be found [here](../elasticsearch.md#rules).
+
+* Test settings are stored in the memory of the node that handled the POST tests settings request.
+Impersonation support will be limited to this node.
+
+* Sometimes is impossible to fetch usernames defined in the test settings.
+If a `users` rule contains a username pattern with a wildcard, to impersonate a user matching a given pattern, you need to enter the username manually.
```
sscarduzio, line 171: full example needed

elasticsearch-details/impersonation.md
```diff
@@ -151,4 +154,18 @@ After clicking add/edit user buttons(5), you will see a dialog with an option to
...
+
+## Impersonation limitations
+
+Impersonation mode has some limitations. Please check if they have an impact on your use cases:
+
+* Not all features available in the ROR configuration are testable with impersonation mode.
+Some rules used in ROR ACL do not support impersonation. For example, auth rule with hashed credentials(e.g. `auth_key_sha512`) can be used in impersonation mode only when credentials follow the format `USER_NAME: HASH(PASSWORD)`; A fully hashed username and password don't allow fetching a username. The auth rule in such a format won't match during impersonation.
+A list of the rules with info about impersonation support can be found [here](../elasticsearch.md#rules).
+
+* Test settings are stored in the memory of the node that handled the POST tests settings request.
```
sscarduzio, line 167: hmmm... what if you define multiple ES nodes in Kibana? Then the experience will be inconsistent?

mateuszkp96: Yes, it looks like the experience will be inconsistent. In the next tasks, we will introduce distributed test settings. But for now, how to inform users about that? I think it's worth informing them about that.

sscarduzio: Yes absolutely, it's a very important detail. Btw I remember now we already talked about this 👍

coutoPL:
Thanks guys. I'm going to modify it a little bit. I will show you tomorrow

kibana/README.md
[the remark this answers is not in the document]
coutoPL: for me it's ok

coutoPL:
@Dzuming @sscarduzio @mateuszkp96 
Guys, I added some context (aka story) to the documentation of the feature. WDYT? 
Do you want to add/change sth? Would a clear enough for someone who is not familiar with sth like impersonation?

@Dzuming maybe you could take a look once again at the examples with screens. Do you think they are used correctly? Maybe we need to change/add sth also here?

coutoPL (changes requested):
@Dzuming @sscarduzio @mateuszkp96
Guys, I added some context (aka story) to the documentation of the feature. WDYT?
Do you want to add/change sth? Would a clear enough for someone who is not familiar with sth like impersonation?

@Dzuming maybe you could take a look once again at the examples with screens. Do you think they are used correctly? Maybe we need to change/add sth also here?

kibana-details/impersonation.md
```diff
@@ -0,0 +1,137 @@
+# Impersonation 
+
+After describing what [the impersonation is](../kibana.md#impersonate), it's high time to see how does ROR support it and who and when could be interested in using this feature. Let's start with the latter one.
+
+## Use cases
+
+The impersonation feature is prepared for a ROR admin. It's not rather useful for users (at least not directly). We can point out two the most obvious use cases when the admin could take advantage of the feature:
```
sscarduzio, line 7: -> The impersonation feature is intended for ROR administrators, rather than users.

kibana-details/impersonation.md
```diff
@@ -0,0 +1,137 @@
...
+
+#### Debugging users' problems:
+
+Let's imagine that some user has a problem with its ROR configuration (eg. the user doesn't has an access to some feature that was blocked at ROR's level by you, the admin). They are not able to clearly describe what the issue is (sounds familiar?). For the admin, it'd be better if they can see what the user sees. Thanks to impersonation feature, an admin is allowed to impersonate the user and feel and see the same the user sees.
```
sscarduzio, line 11: the user doesn't have access

.., and they are not able to clearly describe

As an administrator, it would be extremely beneficial  if you could see what he user sees.

kibana-details/impersonation.md
```diff
@@ -0,0 +1,137 @@
...
+
+#### Debugging users' problems:
+
+Let's imagine that some user has a problem with its ROR configuration (eg. the user doesn't has an access to some feature that was blocked at ROR's level by you, the admin). They are not able to clearly describe what the issue is (sounds familiar?). For the admin, it'd be better if they can see what the user sees. Thanks to impersonation feature, an admin is allowed to impersonate the user and feel and see the same the user sees. 
+
+#### Configuring a new user:
+
+When an admin configures a new user in ROR settings, they face two problems:
+
+1. `Won't the updated configuration break the production cluster?`
```
sscarduzio, line 17: Will the...

kibana-details/impersonation.md
```diff
@@ -0,0 +1,137 @@
...
+1. `How do I know that the new user is correctly configured? Do they have correct permissions?`
```
sscarduzio, line 18: Did I configure all their permission correctly?

kibana-details/impersonation.md
```diff
@@ -0,0 +1,137 @@
...
+
+Both of the problems can be solved using the ROR's impersonation. Thanks to the fact that the impersonation operates on different than production (main) settings, the admin can alter it without worries that their actions will break something and users won't be able to do their job.
```
sscarduzio, line 20: Thanks to the fact that the impersonation feature always uses its own "test settings", that is completely independent from the main production settings

kibana-details/impersonation.md
```diff
@@ -0,0 +1,137 @@
...
+#### Configuring a new user:
+
+When an admin configures a new user in ROR settings, they face two problems:
+
+1. `Won't the updated configuration break the production cluster?`
+1. `How do I know that the new user is correctly configured? Do they have correct permissions?`
+
+Both of the problems can be solved using the ROR's impersonation. Thanks to the fact that the impersonation operates on different than production (main) settings, the admin can alter it without worries that their actions will break something and users won't be able to do their job. 
+
+Admin can add the new user configuration without worries and then test it by impersonate the user. They can check if the user can log in without problems and if the user has access only to the Kibana's features the admin wanted to grant. When the admin is sure that everything is configured correctly, they can promote the settings (test) to production.
```
sscarduzio, line 22: Kibana features

kibana-details/impersonation.md
```diff
@@ -0,0 +1,137 @@
...
+
+Before an admin will be able to impersonate the user, they have to configure ROR properly. The configuration consists of several parts:
+
+1. creating ROR's Test Settings,
+2. defining mocks of the external services,
+3. impersonating a chosen user.
+
+#### Creating ROR's Test Settings
+
+When you call Elasticsearch directly or through ROR Kibana, ROR ACL is defined by Settings (we can assume they are main settings). The Test Settings define another ACL, that is taken into consideration by ROR ES only then a proper impersonation header is passed. The Test Settings are active only for a strictly defined amount of time. After the time is expired, they are automatically invalidated (for a security reasons).
```
sscarduzio, line 34: only ~~then~~ when a proper impersonation header

sscarduzio: for ~~a~~ security reasons

kibana-details/impersonation.md
```diff
@@ -0,0 +1,137 @@
...
+2. defining mocks of the external services,
+3. impersonating a chosen user.
+
+#### Creating ROR's Test Settings
+
+When you call Elasticsearch directly or through ROR Kibana, ROR ACL is defined by Settings (we can assume they are main settings). The Test Settings define another ACL, that is taken into consideration by ROR ES only then a proper impersonation header is passed. The Test Settings are active only for a strictly defined amount of time. After the time is expired, they are automatically invalidated (for a security reasons). 
+
+Detailed instruction how to define Test Settings can be find [here](../examples/impersonation/test-settings-ui.md).
+
+Copying Main Settings as Test Settings is not enough. We have to instruct ROR which users can be considered as impersonators (the ones, who are allowed to impersonate any other user). The list of allowed impersonators and their permissions are defined in the `impersonation` section in ROR Settings:
```
sscarduzio, line 38: We also have to instruct

kibana-details/impersonation.md
```diff
@@ -0,0 +1,137 @@
...
+
+When an impersonator passes wrong credentials ROR will tell Kibana that impersonation is not allowed.
+
+#### Defining mocks of the external services (optional)
+
+ROR has many sophisticated authentication & authorization methods. Some of them are based on external systems like LDAP. A problem with such systems according to the impersonation feature is that, the systems don't support it by default, or the configuration is difficult, or they don't support it at all. 
+
+That's why we decided to solve it totally differently - using mocks. [Wikipedia](https://en.wiktionary.org/wiki/mock) defines `mock` as `an imitation, usually of lesser quality.` And in case of external authentication system we are going provide an imitation of it that tell us what users can be authenticated by it. When we consider and authorization service, a mock of it will return us users with their roles in the service. And this is enough for ROR to support impersonation.
```
sscarduzio, line 66: tell us -> tell the ACL ?

return us users  -> return the ACL?

kibana-details/impersonation.md
```diff
@@ -0,0 +1,137 @@
...
+
+ROR has many sophisticated authentication & authorization methods. Some of them are based on external systems like LDAP. A problem with such systems according to the impersonation feature is that, the systems don't support it by default, or the configuration is difficult, or they don't support it at all. 
+
+That's why we decided to solve it totally differently - using mocks. [Wikipedia](https://en.wiktionary.org/wiki/mock) defines `mock` as `an imitation, usually of lesser quality.` And in case of external authentication system we are going provide an imitation of it that tell us what users can be authenticated by it. When we consider and authorization service, a mock of it will return us users with their roles in the service. And this is enough for ROR to support impersonation. 
+
+How does ROR use the mocks? Let's suppose we have `ldap_auth` rule. When ROR processes the rule, it:
+* asks a given LDAP if the username can be authenticated with a given password, and if they can ...
+* asks the LDAP to tell what groups the user has
+
+In the impersonation case, it looks pretty the same. The only difference is that ROR not calling the LDAP but the LDAP's mock and it:
```
sscarduzio, line 72: * asks LDAP to list what groups the user belongs to

In the impersonation case, it looks pretty much the same. Just we won't call any LDAP server, rather, the mock will provide the required information (no passwords required).

kibana-details/impersonation.md
```diff
@@ -0,0 +1,137 @@
...
+
+Now, when we have configured Test Settings and External Services Mocks, we can try to impersonate a user. In ROR Settings, user can be:
+* provided statically (defined in the settings),
+* provided dynamically:
+   * from external, dependant systems (like LDAP) - we mock them
+   * from upstream systems (eg. through headers) - they are unknown
+
+It means that we pick the users defined in Settings or Mocks, but also we can enter the username and try to impersonate such user.
+
+[Here](../examples/impersonation/impersonate-user-ui.md) you find a description how to impersonate an user in ROR Kibana.
```
sscarduzio, line 90: an  user -> a user

sscarduzio (commented):
When documentation talks about the user, and not about the product, it's much better documentation. Thanks!

I took the time to suggests some edits!

mateuszkp96 (commented):
LGTM! Use cases are described well, and what's most important, you show why I (as administrator) should use impersonation.
One comment from my side left.

kibana.md
```diff
@@ -1001,3 +1001,17 @@ readonlyrest:
 ```
 
 Now try to login as user1, and ReadonlyREST Enterprise should initialise the index ".kibana\_user1" with all the index patterns and dashboards contained in the template tenancy.
+
+## Impersonation
```
Dzuming, line 1005: Shouldn't we add somehow keywords like test settings, and auth mock service into this description?  When the user opens the UI and will see these kinds of words,  and then, will try to search for them, he won't be able to find this description and go into the GitHub page

Dzuming:
@coutoPL Updated a few images, the rest looks ok :)

Dzuming:
> Guys, are we ready to ask Janek to read it though and tell us what he (as someone who is not familiar with the topic, so not biased) thinks?

Let's do it

coutoPL (approved):

sscarduzio:
I invited @here-s-johnny as a maintainer, he has to accept-

details/impersonation.md
```diff
@@ -0,0 +1,144 @@
...
+
+How does ROR use the mocks? Let's suppose we have `ldap_auth` rule. When ROR processes the rule, it:
+* asks a given LDAP if the username can be authenticated with a given password, and if they can ...
+* asks LDAP to list what groups the user belongs to
+
+In the impersonation case, it looks pretty much the same. Just we won't call any LDAP server, rather, the mock will provide the required information (no passwords required):
+* asks the mock if the username exists, and if it does ...
+* asks the mock to tell what groups the user belongs to
+
+**⚠️ IMPORTANT:** When some external service is not mocked, ROR might return to Kibana information that the impersonation is not supported. It's better to always define all mocks, to avoid the "not supported impersonation" ES response.
```
here-s-johnny, line 76: - `If one or more of the external services are not mocked, ROR might inform Kibana that the impersonation is not supported.`
- `... to avoid the "Impersonation not supported" ES response`

details/impersonation.md
```diff
@@ -0,0 +1,144 @@
...
+
+How to define mocks in Kibana is described [here](../examples/impersonation/external-services-mocks-ui.md).
+
+#### Impersonating a chosen user
+
+Now, when we have configured Test Settings and External Services Mocks, we can try to impersonate a user. In ROR Settings, user can be:
```
here-s-johnny, line 82: - `Now that we have configured...`
- "In ROR Settings" meaning readonlyrest.yml or kibana.yml or both? 🤔

details/impersonation.md
```diff
@@ -0,0 +1,144 @@
...
+* provided statically (defined in the settings),
+* provided dynamically:
+   * from external, dependant systems (like LDAP) - we mock them
+   * from upstream systems (eg. through headers) - they are unknown
```
here-s-johnny, line 86: I don't understand what "they are unknown" means 🤔 Are they not known upfront? Or not at all?

details/impersonation.md
```diff
@@ -0,0 +1,144 @@
...
+
+It means that we pick the users defined in Settings or Mocks, but also we can enter the username and try to impersonate such user.
```
here-s-johnny, line 88: This is followup to the previous comment and maybe it stems from my limited knowledge of how ROR_ES works, but I don't understand this. Who are these users not defined in settings nor in mocks? 🤔

details/impersonation.md
```diff
@@ -0,0 +1,144 @@
...
+
+[Here](../examples/impersonation/impersonate-user-ui.md) you find a description how to impersonate a user in ROR Kibana.
```
here-s-johnny, line 90: `... you can find...`

details/impersonation.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Logs & audit
+
+In ES logs in `USR` admin is going to find sth like this: `admin1 as (user1)` - it means that `admin1` was authenticated and they are the impersonator who is impersonating `user1`.
```
here-s-johnny, line 94: - `In ES logs if an admin user ('USR') finds something like this:...`
- I think especially the sth -> something is important, as it an official public documentation, it should be tidy

details/impersonation.md
```diff
@@ -0,0 +1,144 @@
...
+All logs of impersonated user in Kibana will have this format `[<log level>][plugins][ReadonlyREST][<filename>][impersonating <impersonated user username>]`
+
+When auditing is enabled, the audit document is going to contain an `impersonated_by` field.
+
+## Impersonation limitations
+
+Impersonation mode has some limitations. Please check if they have an impact on your use cases:
+
+* Not all features available in the ROR configuration are testable with impersonation mode.
+Some rules used in ROR ACL do not support impersonation. For example, auth rule with hashed credentials(e.g. `auth_key_sha512`) can be used in impersonation mode only when credentials follow the format `USER_NAME: HASH(PASSWORD)`; A fully hashed username and password don't allow fetching a username. The auth rule in such a format won't match during impersonation.
```
here-s-johnny, line 105: Space missing before the bracket 👾

details/impersonation.md
```diff
@@ -0,0 +1,144 @@
...
+A list of the rules with info about impersonation support can be found [here](../elasticsearch.md#rules).
+
+* Test settings are stored in the memory of the node that handled the POST tests settings request.
```
here-s-johnny, line 108: - Test Settings should be capitalised all through the document, for consistency's sake
- `...handled the POST Test Settings request`

details/impersonation.md
```diff
@@ -0,0 +1,144 @@
...
+Impersonation support will be limited to this node. We are going to improve it in the future, but for now your kibana should talk only to one ES node.
```
here-s-johnny, line 109: `... your Kibana should only communicate with one ES node`

details/impersonation.md
```diff
@@ -0,0 +1,144 @@
...
+
+* Sometimes is impossible to fetch usernames defined in the test settings.
```
here-s-johnny, line 111: - `Sometimes it is impossible...`
- `...in the Test Settings`

details/impersonation.md
```diff
@@ -0,0 +1,144 @@
...
+If a `users` rule contains a username pattern with a wildcard, to impersonate a user matching a given pattern, you need to enter the username manually.
```
here-s-johnny, line 112: `... to impersonate a user matching the pattern...`

details/impersonation.md
```diff
@@ -0,0 +1,144 @@
...
+        
+    impersonation:
+      [...]
+  ```
+
+## Glossary
+
+* **Impersonator** - someone who imitates or copies the behavior or actions of another,
+* **Impersonation** - imitating behaviors or actions of a given user,
+* **Main Settings** - the ROR's settings that apply to ACL that handles request during regular sessions (not the impersonation ones),
```
here-s-johnny, line 142: `... that handles requests...`

examples/impersonation/impersonate-user-ui.md
```diff
@@ -0,0 +1,20 @@
...
+    ![Impersonate ror menu](<../../.gitbook/assets/test_settings_ror_menu.png>)
+
+3. Go into the Impersonate tab
+4. You can free type impersonate. This button is available only in the situation when in some cases, the system is not able to receive all usernames. In this case, to impersonate, you need to type impersonating username manually.
+5. You can add/edit user in a specific external auth mock service
+6. You can impersonate a user and imitate his behavior and actions
+
+    ![Impersonate tab](<../../.gitbook/assets/impersonate_tab.png>)
+
+7. When an impersonation session is started correctly, the "impersonating" will be visible in the ROR menu a shown in the picture.
```
here-s-johnny, line 16: `... as shown...`

examples/impersonation/test-settings-ui.md
```diff
@@ -0,0 +1,16 @@
...
+
+For impersonation to work, some valid Test Settings should be created and saved. It's important that the main ROR setting will be unaffected, so you, as an admin/user don't need to worry that you will break something. Here is how to write test settings:
+
+1. Open the ROR menu
+2. Click the Edit security settings button
+
+    ![Test settings ror menu](<../../.gitbook/assets/test_settings_ror_menu.png>)
+
+1. Go into the Test settings tab
+2. You can set the "time to live" (TTL), that is a time interval after which the test settings will be automatically deactivated and impersonation session will also abruptly exit
```
here-s-johnny, line 11: Are the numbers ok here?

kibana.md
```diff
@@ -1001,3 +1001,17 @@ readonlyrest:
...
+
+According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
+
+> An impersonator is someone who imitates or copies the behavior or actions of another.
+
+So, an impersonation can be understood as imitating behaviors or actions.
+In the context of ReadonlyREST: one user could imitate an action 
+of another user. Why would we want it? Let's suppose the first user is 
+an admin, who has just configured access for a new user. They would like 
+to know if the rule(s) are configured correctly. And here it comes the impersonation feature. The admin can impersonate given user in Kibana and see what the user would see if they logged in themselves.
```
here-s-johnny, line 1015: `... here comes the impersonation feature...`

details/impersonation.md
```diff
@@ -0,0 +1,144 @@
...
+#### Debugging users' problems:
+
+Let's imagine that some user has a problem with its ROR configuration (eg. the user doesn't have an access to some feature that was blocked at ROR's level by you, the admin). And they are not able to clearly describe what the issue is (sounds familiar?). As an administrator, it would be extremely beneficial if you could see what he user sees. Thanks to impersonation feature, an admin is allowed to impersonate the user and feel and see the same the user sees. 
+
+#### Configuring a new user:
+
+When an admin configures a new user in ROR settings, they face two problems:
+
+1. `Will the updated configuration break the production cluster?`
+1. `How do I know that the new user is correctly configured? Did I configure all their permission correctly??`
```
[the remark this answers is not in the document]
coutoPL: the doubled 1. - it's a feature ;) (see https://www.markdownguide.org/basic-syntax/)

details/impersonation.md
coutoPL: In ROR settings, in general, a user can be:
* defined statically (in some cases ROR doesn't know the usernames because eg. they are hashed)
* be defined in an external service (eg. LDAP)
* be sent to ROR (with JWT or header)

IMO there is no sense to write here about it because it's rather related to the ROR rules.

examples/impersonation/test-settings-ui.md
coutoPL: the same markdown feature ;)

sscarduzio:
Hi guys, sorry being late, but reading this again I found [a repeated anti pattern](https://www.w3.org/QA/Tips/noClickHere) we can easily fix.

details/impersonation.md
```diff
@@ -0,0 +1,144 @@
...
+* asks the given LDAP service if the username can be authenticated with a given password, and if they can ...
+* asks LDAP to list what groups the user belongs to
+
+In the impersonation case, it looks pretty much the same. The difference being that ROR won't call any LDAP server - the mock will provide the required information instead (no password required). During impersonating, when ROR processes an LDAP rule, it:
+* asks the mock if the username exists, and if it does ...
+* asks the mock to tell what groups the user belongs to
+
+**⚠️ IMPORTANT:** If one or more of the external services are not mocked, ROR might inform Kibana that the impersonation is not supported. It's better to always define all mocks, to avoid the "Impersonation not supported" Elasticsearch response.
+
+How to define mocks in Kibana is described [here](../examples/impersonation/external-services-mocks-ui.md).
```
sscarduzio, line 78: https://www.w3.org/QA/Tips/noClickHere

details/impersonation.md
```diff
@@ -0,0 +1,144 @@
...
+
+Now that we have configured Test Settings and External Services Mocks, we can try to impersonate a user. In Elasticsearch ROR Settings, user can be:
+* provided statically (defined in the settings),
+* provided dynamically:
+   * from external, dependant systems (like LDAP) - we mock them
+   * from upstream systems (eg. through headers) - they are not known upfront
+
+It means that we pick the users defined in Settings or Mocks, but also we can enter the username and try to impersonate such user.
+
+[Here](../examples/impersonation/impersonate-user-ui.md) you can find a description how to impersonate a user in ROR Kibana.
```
sscarduzio, line 90: https://www.w3.org/QA/Tips/noClickHere

details/impersonation.md
```diff
@@ -0,0 +1,144 @@
...
+
+When auditing is enabled, the audit document is going to contain an `impersonated_by` field.
+
+## Impersonation limitations
+
+Impersonation mode has some limitations. Please check if they have an impact on your use cases:
+
+* Not all features available in the ROR configuration are testable with impersonation mode.
+Some rules used in ROR ACL do not support impersonation. For example, auth rule with hashed credentials (e.g. `auth_key_sha512`) can be used in impersonation mode only when credentials follow the format `USER_NAME: HASH(PASSWORD)`; A fully hashed username and password don't allow fetching a username. The auth rule in such a format won't match during impersonation.
+A list of the rules with info about impersonation support can be found [here](../elasticsearch.md#rules).
```
sscarduzio, line 106: https://www.w3.org/QA/Tips/noClickHere

kibana.md
```diff
@@ -1001,3 +1001,17 @@ readonlyrest:
...
+
+> An impersonator is someone who imitates or copies the behavior or actions of another.
+
+So, an impersonation can be understood as imitating behaviors or actions.
+In the context of ReadonlyREST: one user could imitate an action 
+of another user. Why would we want it? Let's suppose the first user is 
+an admin, who has just configured access for a new user. They would like 
+to know if the rule(s) are configured correctly. And here comes the impersonation feature. The admin can impersonate given user in Kibana and see what the user would see if they logged in themselves. 
+
+You can check more detailed information about impersonation feature [here](./details/impersonation.md)
```
sscarduzio, line 1017: https://www.w3.org/QA/Tips/noClickHere

details/impersonation.md
```diff
@@ -0,0 +1,144 @@
...
+
+1. creating ROR's Test Settings,
+2. defining mocks of the external services (like [LDAP](../elasticsearch.md#ldap-connector), [External Basic Auth](../elasticsearch.md#external-basic-auth) or [Custom groups provider](../elasticsearch.md#custom-groups-providers)),
+3. impersonating a chosen user.
+
+#### Creating ROR's Test Settings
+
+When you call Elasticsearch directly or through ROR Kibana, ROR ACL is defined by Settings (we can assume they are Main Settings). The Test Settings define another ACL, that is taken into consideration by ROR ES only when a proper impersonation header is passed. The header is managed by ROR internally. The Test Settings are active only for a strictly defined amount of time (by default it's *30 minutes*, but the admin can change it before applying Test Settings). After the time has expired, they are automatically invalidated (for security reasons). Obviously, the admin is allowed to invalidate the configured Test Settings in any time. There is no way to have more than one Test Settings configured at time.
+
+Detailed instructions on how to define Test Settings can be found [here](../examples/impersonation/test-settings-ui.md).
```
sscarduzio, line 36: https://www.w3.org/QA/Tips/noClickHere

sscarduzio (changes requested):

details/impersonation.md
coutoPL: I'm not sure if I "feel" the rule. But I've changed all the places with the "here link". Please let me know if it's ok now or maybe you have an idea on how to formulate it in a better way

sscarduzio: The clickable text on a hyperlink should give sufficient context to the reader when they scan a paragraph. When the link text says "click here", the user is forced to rebuild the context from the sentences around the actual link instead, and the whole UX deteriorates.

When you go through documentation especially, you are not reading sequentially. You are scanning for information because you have a problem to solve. So if we are forced to hide information inside a different page, the user at least needs to know quickly what's behind the link before they decide to invest into changing page (and losing the current context).

details/impersonation.md
```diff
@@ -33,9 +33,11 @@ Before an admin will be able to impersonate a user, they have to configure ROR p
 
 When you call Elasticsearch directly or through ROR Kibana, ROR ACL is defined by Settings (we can assume they are Main Settings). The Test Settings define another ACL, that is taken into consideration by ROR ES only when a proper impersonation header is passed. The header is managed by ROR internally. The Test Settings are active only for a strictly defined amount of time (by default it's *30 minutes*, but the admin can change it before applying Test Settings). After the time has expired, they are automatically invalidated (for security reasons). Obviously, the admin is allowed to invalidate the configured Test Settings in any time. There is no way to have more than one Test Settings configured at time.
 
-Detailed instructions on how to define Test Settings can be found [here](../examples/impersonation/test-settings-ui.md).
+ROR Kibana plugin provides dedicated UI to easily manage Test Settings. The [detailed instruction](../examples/impersonation/test-settings-ui.md) describes how to do that.
```
sscarduzio, line 36: The idea is to avoid pronouns. We could simplify  the above text as such:

See how to  [manage Test Settings using ROR Kibana plugin](...) with the dedicated UI.

details/impersonation.md
```diff
@@ -75,7 +77,7 @@ In the impersonation case, it looks pretty much the same. The difference being t
 
 **⚠️ IMPORTANT:** If one or more of the external services are not mocked, ROR might inform Kibana that the impersonation is not supported. It's better to always define all mocks, to avoid the "Impersonation not supported" Elasticsearch response.
 
-How to define mocks in Kibana is described [here](../examples/impersonation/external-services-mocks-ui.md).
+ROR Kibana plugin obviously helps admins to define mocks in the easy way. The [detailed instruction](../examples/impersonation/external-services-mocks-ui.md) describes how to do that.
```
sscarduzio, line 80: Another good idea in docs is to stay away from adverbs like obviously, simply, just, easily, used in absolute way:

❌ "to accomplish XYZ, <simply|just|obviously> do this" <-  (makes assumptions of what SHOULD be easy for the reader: it's judgemental and not informative).

✅ "An easier way to accomplish YQW would be using this tool..."   <-- (something is easier than something else, OK) 

Would rewrite this to:

"Follow the detailed instructions on how admins can [define  and configure services mocks]() using the dedicated graphical user interface".

details/impersonation.md
```diff
@@ -87,7 +89,7 @@ Now that we have configured Test Settings and External Services Mocks, we can tr
 
 It means that we pick the users defined in Settings or Mocks, but also we can enter the username and try to impersonate such user.
 
-[Here](../examples/impersonation/impersonate-user-ui.md) you can find a description how to impersonate a user in ROR Kibana.
+The [step by step instruction](../examples/impersonation/impersonate-user-ui.md) shows on how to impersonate a user using ROR Kibana plugin UI.
```
sscarduzio, line 92: BTW Instructions used like this need to be plural.

"Follow the instructions on how to [impersonate a user using the ROR Kibana plugin UI](...)."

details/impersonation.md
coutoPL: the same violation here (as I mentioned, for me it shows that [the guide](https://www.w3.org/QA/Tips/noClickHere) at least is not intuitive)

details/impersonation.md
sscarduzio: Yeah good point! This should be better:

"ROR Kibana plugin provides a dedicated Test Settings UI. See our [Test Settings management guide](....) for more information."

details/impersonation.md
sscarduzio: Yep.

"ROR Kibana plugin helps administrators to visually create and edit service mocks with a dedicated graphical UI. Follow our [service mock configuration guide](.....) for more. "

PR #115, opened 2022-05-26
RORDEV-671-assets-update

Dzuming:
Reduce size of all assests. See https://github.com/beshu-tech/readonlyrest-docs/pull/114#issuecomment-1138160136

I moved it into separate pr because, well... it kills the browser because of amount of changes


Dzuming:
yes, I think I will merge it directly into the develop, or when all comments related to documentation will be resolved (I assume I will be the one who will click merge into develop, since I'm a creator of impersonation docs pr :D ). Because if I merge it, then it won't be possible to check file changes, Github won't be able to render these all changes

coutoPL:
@Dzuming could you please take a look at the merge conflicts?

PR #118, opened 2022-06-11
RORDEV-665 'roles_and' in 'ror_kbn_auth' and 'jwt_auth' rules

mateuszkp96:
Provided `groups_and` description for `ror_kbn_auth` and `jwt_auth` rules

### DIFF
--- a/elasticsearch.md
+++ b/elasticsearch.md
@@ -1120,7 +1120,7 @@ readonlyrest:
       users: ["root", "*@mydomain.com"]
       jwt_auth:
         name: "jwt_provider_1"
-        roles: ["viewer"]
+        groups: ["viewer"]
 ```
 
 #### `groups`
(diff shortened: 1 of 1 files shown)


elasticsearch.md
```diff
@@ -1274,7 +1274,12 @@ readonlyrest:
     - name: "ReadonlyREST Enterprise instance #1"
       ror_kbn_auth:
         name: "kbn1"
-        roles: ["SAML_GRP_1"]
+        roles: ["SAML_GRP_1", "SAML_GRP_2"] # <-  use this rule when a user should have one of the listed roles
```
coutoPL, line 1277: in other authorization rules, we have the `groups` field. In JWT-based rules, we maintain `roles` for the backward comparibility, but for sake of consistency, `groups` was introduced here some time ago. So, let's have `groups` in examples. And let's don't use the `roles` word when we write about groups.

coutoPL: `use this rule when a user should` ... `roles|groups` is a field or property of the `ror_kbn_auth` rule

coutoPL (changes requested):

coutoPL (approved):

PR #120, opened 2022-07-21
[RORDEV-699] Patching description improvements

coutoPL:


elasticsearch.md
```diff
@@ -64,15 +64,33 @@ Notice how we need to type in the format `file://` + absolute path \(yes, with t
...
+You can verify if Elasticsearch was correctly patched using command `verify`:
+
+```bash
+$ jdk/bin/java -jar plugins/readonlyrest/ror-tools.jar verify 
+```
+
+Please note that the tool assumes that the default installation directory is `/usr/share/elasticsearch`. You can instruct it to use different directory by executing one of tool's command with `--es-path` parameter:
+
+```bash
+$ jdk/bin/java -jar plugins/readonlyrest/ror-tools.jar patch --es-path /my/custom/path/to/es/folder
```
sscarduzio, line 87: Could we put in the examples a typical, realistic es path? I.e. default Linux es paths. 

One more thing: I'd try to predict 2 or 3 typical paths before giving up and requiring the user to enter the path via the cli option.

coutoPL: I've checked all the installation options from there: https://www.elastic.co/guide/en/elasticsearch/reference/current/install-elasticsearch.html

Always installation path is `/usr/share/elasticsearch`. But you can download the zip file, unpack and put it where you want. We are able to detect if some directory looks like ES one, so we could try to find it in the user's filesystem, but I don't think it's a good idea.

But I improved the tool according to your suggestion. Now, I'm checking PWD and then trying to fallback to /usr/share/elasticsearch. If you run the tool from the root of the ES directory it's going to work no matter where it's located

coutoPL: and as you can see that our examples suggest that the user is running the tool from the root of the ES directory

coutoPL: @sscarduzio do you have any other comments?

PR #122, opened 2022-08-09
RORDEV-715 Provide info about new kibana_custom_js_inject_file kibana.yml property

Dzuming:
- Provide info about new kibana_custom_js_inject_file kibana.yml property

### DIFF
--- a/kibana.md
+++ b/kibana.md
@@ -889,6 +889,12 @@ readonlyrest_kbn.kibana_custom_css_inject_file: '/tmp/custom.css'
 readonlyrest_kbn.kibana_custom_js_inject: '$(".global-nav__logo").hide(); alert("hello!")'
 ```
 
+Alternatively, it's possible to load the JS from a file in the filesystem:
+
+```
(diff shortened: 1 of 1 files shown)


kibana.md
```diff
@@ -889,6 +889,12 @@ readonlyrest_kbn.kibana_custom_css_inject_file: '/tmp/custom.css'
 readonlyrest_kbn.kibana_custom_js_inject: '$(".global-nav__logo").hide(); alert("hello!")'
 ```
 
+Alternatively, it's possible to load the JS from a file in the filesystem:
+
+```
```
sscarduzio, line 894: Worth mentioning for this new property (and its CSS counterpart) that if you use relative paths, you end up pointing to kibana home, i.e. 
```
readonlyrest_kbn.kibana_custom_js_inject: 'config/custom.js'
```
Will refer to `$KBN_HOME/config/custom.js` which is the same directory where `kibana.yml` can normally be found.

Dzuming: updated ✔️

PR #125, opened 2022-09-20
Move validate inside http_connection_settings.

tristal:
Sadly vscode issued an autoformat on the file, which isn't a bad thing, but caused a bunch of whitespace cleanup.  If you would prefer this to not happen, I can force it to not occur.

The only change to the text is moving the `validate: false # SSL certificate validation (default to true)` inside of the `http_connection_settings` element, as when trying to add this to my own configuration it fails the RoR startup validation.


coutoPL (approved):
Thanks :)

PR #126, opened 2022-09-20
[RORDEV-731] dynamic variables docs improvements + Internode SSL new settings

coutoPL:
ref: https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/846

### DIFF
--- a/elasticsearch.md
+++ b/elasticsearch.md
@@ -1760,14 +1760,50 @@ One of the neatest feature in ReadonlyREST is that you can use dynamic variables
 
 You can let users authenticate externally, i.e. via LDAP, and use their user name string inside the `indices` rule.
 
-```text
+```yaml
 readonlyrest:
     access_control_rules:
 
-    - name: "Users can see only their logstash indices i.e. alice can see alice_logstash-20170922"
+    - name: "Users can see only their logstash indices" # i.e. alice can see alice_logstash-20170922
       ldap_authentication:
         name: "myLDAP"
-      indices: ["@{acl:user}_logstash-*"]
+      indices: ["@{acl:user}_logstash-*"] 
+
+    # LDAP connector settings omitted, see LDAP section below..
+```
+
+#### Indices from available groups
+
+You can let users authorize externally, i.e. via LDAP, and use their group names strings inside the `indices` rule.
+
+```yaml
+readonlyrest:
+    access_control_rules:
+
+    - name: "Users can see only logstash indices for their departments" # i.e. alice belongs to 'dev' and 'ops' department groups, so she can see dev_logstash-20170922, ops_logstash-20170922
+      ldap_auth:
+        name: "myLDAP"
+        groups: ["dev", "ops", "qa"]
+      indices: ["@explode{acl:available_groups}_logstash-*"] #  notice that 'available_groups' is an array. `indices` rule has multi-value context (we can configure an array of index names in it). It means that we have to create several indices names from the available_groups array and the optional static part (here: _logstash-*) - it can be achieved with "explode" keyword i.e from available_groups=[dev, ops] we will get: indices: ["dev_logstash-*", "ops_logstash-*"] 
+
+    # LDAP connector settings omitted, see LDAP section below..
+```
+
+#### Filter from available groups
+
+You can let users authorize externally, i.e. via LDAP, and use their group names strings inside the `filter` rule.
+
+
+```yaml
+readonlyrest:
+    access_control_rules:
+
+    - name: "Users can only see documents related to their departments" # i.e. alice belongs to 'dev' and 'ops' department groups, so she can see documents where "department" field is equal 'dev' or 'ops' 
+      ldap_auth:
+        name: "myLDAP"
+        groups: ["dev", "ops", "qa"]
+      filter: '{ "terms": { "department": [@{acl:available_groups}] }}' # here we don't use `explode` keyword, because it's a context of single value. `available_groups` array values will be surrounded by double quotes and contatenated with comma i.e. from availabe_groups=[dev, ops] we will get: filter: '{ "terms": { "department": ["dev","ops"] }}'
+      indices: ["logstash-*"]
 
     # LDAP connector settings omitted, see LDAP section below..
 ```
@@ -1776,7 +1812,7 @@ readonlyrest:
 
 You can let users authorize externally, i.e. via LDAP, and use their group inside the `uri_re` rule.
 
-```text
+```yaml
 readonlyrest:
     access_control_rules:
 
@@ -1793,7 +1829,7 @@ readonlyrest:
 
 Imagine that we delegate authentication to a reverse proxy, so we know that only authenticated users will ever reach Elasticsearch. We can tell the reverse proxy \(i.e. Nginx\) to inject a header called `x-nginx-user` containing the username.
 
-```text
+```yaml
 readonlyrest:
     access_control_rules:
 
--- a/elasticsearch.md
+++ b/elasticsearch.md
@@ -434,6 +434,7 @@ readonlyrest:
     key_pass: [ password for generated certificate ]
     truststore_file: "elastic-certificates.p12"
     truststore_pass: [ password for generated certificate ]
+    client_authentication: true # default: false
     certificate_verification: true # certificate verification is enabled by default on XPack nodes
 ```
 
@@ -446,7 +447,7 @@ It is useful on local/test environment, where security is not the most important
 certificate_verification: true
 ```
 
-under `ssl_internode` section. This option is applicable only for internode ssl.
+under `ssl_internode` section. This option is applicable only for internode SSL.
 
 #### Client authentication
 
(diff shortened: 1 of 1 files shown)


elasticsearch.md
```diff
@@ -1752,22 +1751,59 @@ And ReadonlyREST ES will load "S3cr3tP4ss" as `bind_password`.
...
+
+```yaml
+readonlyrest:
+    access_control_rules:
+
+    - name: "Users can see only logstash indices for their departments" # i.e. alice belongs to 'dev' and 'ops' department groups, so she can see dev_logstash-20170922, ops_logstash-20170922
+      ldap_auth:
+        name: "myLDAP"
+        groups: ["dev", "ops", "qa"]
+      indices: ["@explode{acl:available_groups}_logstash-*"] #  notice that 'available_groups' is an array. `indices` rule has multi-value context (we can configure an array of index names in it). It means that we have to create several indices names from the available_groups array and the optional static part (here: _logstash-*) - it can be achieved with "explode" keyword i.e from available_groups=[dev, ops] we will get: indices: ["dev_logstash-*", "ops_logstash-*"]
```
sscarduzio, line 1787: woa this in-comment explanation has a ton of detail! How about we create a small paragraph with a few examples, and in the comment we put the "expanded" conf?

elasticsearch.md
```diff
@@ -1752,22 +1751,59 @@ And ReadonlyREST ES will load "S3cr3tP4ss" as `bind_password`.
...
+
+```yaml
+readonlyrest:
+    access_control_rules:
+
+    - name: "Users can only see documents related to their departments" # i.e. alice belongs to 'dev' and 'ops' department groups, so she can see documents where "department" field is equal 'dev' or 'ops' 
+      ldap_auth:
+        name: "myLDAP"
+        groups: ["dev", "ops", "qa"]
+      filter: '{ "terms": { "department": [@{acl:available_groups}] }}' # here we don't use `explode` keyword, because it's a context of single value. `available_groups` array values will be surrounded by double quotes and contatenated with comma i.e. from availabe_groups=[dev, ops] we will get: filter: '{ "terms": { "department": ["dev","ops"] }}'
```
sscarduzio, line 1805: same here!

sscarduzio:
@coutoPL do I remember correctly we also have an alternative to the syntax `${SOME_ENV_VAR}` that goes like: `@{env:SOME_ENV_VAR}`?

elasticsearch.md
```diff
@@ -1719,64 +1718,134 @@ readonlyrest:
...
+One of the neatest features in ReadonlyREST is that you can use dynamic variables inside most values of the following rules: `indices`, `users`, `groups`, `groups_and`, `fields`, `filter`,  `repositories`, `snapshots`, `response_fields`, `uri_re`, `x_forwarded_for`, `hosts_local`, `hosts`, `kibana_index`, `kibana_template_index`. The variables are related to different contexts:
+* `acl` - the context of data collected in authentication and authorization rules of the current block:
+    * `@{acl:user}` gets replaced with the username of the successfully authenticated user. Using this variable is allowed only in blocks where one of the rules is an authentication rule of course it must be rule different from the one containing the given variable.
+    * `@{acl:current_group}` gets replaced with user's current group. Usually resolved by authorization rule defined in a block, but value can be also retrieved by means of kibana plugin. This variable doesn't specify usage requirements.
+    * `@{acl:available_groups}` gets replaced with available groups found in the authorization rule (bacause group strings will be surrounded with double quotes and joined with a comma)
```
sscarduzio, line 1752: available_groups <-- shouldn't this be the list of groups from all groups rules contained in all the ACL blocks that ever match the current request?  Plus, I did not understand the "because" sentence in parentheses 🤔

elasticsearch.md
```diff
@@ -1719,64 +1718,134 @@ readonlyrest:
...
+* `header` - the context of ES HTTP request headers
+    * `@{header:<header_name>}` gets replaced with the value of the HTTP header with name `<header_name>` included in the incoming request \(useful when reverse proxies handle authentication\)
+* `jwt` - the context of JWT passed in the Authorization header
```
sscarduzio, line 1755: the JWT token can be passed in any header, as per ROR configuration.

elasticsearch.md
```diff
@@ -1719,64 +1718,134 @@ readonlyrest:
...
+    * `@{acl:user}` gets replaced with the username of the successfully authenticated user. Using this variable is allowed only in blocks where one of the rules is an authentication rule of course it must be rule different from the one containing the given variable.
+    * `@{acl:current_group}` gets replaced with user's current group. Usually resolved by authorization rule defined in a block, but value can be also retrieved by means of kibana plugin. This variable doesn't specify usage requirements.
+    * `@{acl:available_groups}` gets replaced with available groups found in the authorization rule (bacause group strings will be surrounded with double quotes and joined with a comma)
+* `header` - the context of ES HTTP request headers
+    * `@{header:<header_name>}` gets replaced with the value of the HTTP header with name `<header_name>` included in the incoming request \(useful when reverse proxies handle authentication\)
+* `jwt` - the context of JWT passed in the Authorization header 
+    * `@{jwt:<json_path>}` get replaced with value (or values) found in JWT under the given JSON path
```
sscarduzio, line 1756: "gets replaced with the value (or values) found in the JWT claim under the given JSON path."

elasticsearch.md
```diff
@@ -1719,64 +1718,134 @@ readonlyrest:
...
+#### Dynaminc variables exploding
 
-* `@{acl:user}` gets replaced with the username of the successfully authenticated user. Using this variable is allowed only in blocks where one of the rules is authentication rule \(of course it must be rule different from the one containing given variable\).
-* `@{user}` old style user variable definition. Preferred approach is to use `@{acl:user}`.
-* `@{acl:current_group}` gets replaced with user's current group. Usually resolved by authorization rule defined in block, but value can be also retrieved by means of kibana plugin. This variable doesn't specify usage requirements.
-* `@{xyz}` gets replaced with any `xyz` HTTP header included in the incoming request \(useful when reverse proxies handle authentication\)
+A value resolved from a dynamic variable is a string. Some rules, like `indices` one, have multivalue context (you can configure serveral indices names in it). 
 
-#### Indices from user name
+Let's assume we have a request with the header: `APPS: app1,app2,app3`. Doing something like that:
```
sscarduzio, line 1762: s/that/this/

elasticsearch.md
```diff
@@ -1719,64 +1718,134 @@ readonlyrest:
...
+indices: ["logstash_@{header:apps}"]
+```
+
+We should expect it to be resolved to:
+
+```yaml
+indices: ["logstash_app1,app2,app3"]
+```
+
+for this particular request. I don't think it would be helpful.
```
sscarduzio, line 1774: Avoid first person

elasticsearch.md
```diff
@@ -1719,64 +1718,134 @@ readonlyrest:
...
+But there is an `explode` attribute for dynamic variables. Doing:
```
sscarduzio, line 1775: I'd say explode is more of a function than an attribute, right?

elasticsearch.md
```diff
@@ -1719,64 +1718,134 @@ readonlyrest:
...
+
+we should get:
+
+```yaml
+indices: ["logstash_app1", "logstash_app2", "logstash_app3"]
+```
+
+which looks more useful!
+
+So, as we seen, the `explode` attribute of a dynamic variable rule can be used to split string with comma separated values into an array. But it can only be used in a rule with multivalue context.
```
sscarduzio, line 1789: ..to split a string with comma separated values into an array of strings.

sscarduzio (commented):
Great refactor! Variables are ROR biggest super power and deserved more consideration in our documentation!

I left small comments here and there :)

Thanks!

elasticsearch.md
coutoPL: no, available_groups returned in the response for `/_readonlyrest/metadata/current_user/` is a sum of the set of groups returned by each block. 

`@{acl:available_groups}` var can be resolved to the groups found in the current block only. Groups are collected only by Authorization rules. When we process each block of rules, the processing context is new every time (no gathered data is passed between the processing of the following blocks)

coutoPL: `Plus, I did not understand the "because" sentence in parentheses 🤔` 
heh, me neither :D it looks like I removed some part of the sentence accidentally.

elasticsearch.md
coutoPL: good point

elasticsearch.md
coutoPL: hmm, yes, from the user perspective it's more like a function. 
From the env perspective, it's a marker (or attribute) that changes the way how it's resolved. But the first one seems to be more intuitive.

sscarduzio: In general we should struggle to keep up the blatantly artificial illusion of simplicity, hiding the implementation complexity and handling obvious corner cases and fallbacks under the hood. So the user's mental model stays extra simple. This gives the "it just works" feeling.  

Imagine the difference between setting up "suspend to disk" in Linux (2 hours googling, learn CPU states, swap partitions,...),  and just closing the lid on your Macbook and go by your day. 

MacOS handles all the same things, but they are hidden behind the fluffy curtains of automation and great UX.

Similarly for our case:

DOCUMENTATION=Close the lid, go by your day.
IMPLEMENTATION=Screen brightness: 0%; Suspend to memory for 40 minutes; if still idle after 40m:  if partition does not exist, create a swap file; dump RAM to swap; try set CPU state to S4, if failed, try S3.

Never leak complexity in documentation :)

sscarduzio:
🦄 🕺 😊

PR #128, opened 2022-09-28
Documenting major Kibana version upgrade technique in case of multi tenancy

sscarduzio:


kibana.md
```diff
@@ -181,11 +181,33 @@ $ bin/kibana-plugin remove readonlyrest_kbn
...
+
+The first time you run Kibana after a major version upgrade (e.g. upgrading from Kibana 7.17.7 to Kibana 8.0.0), Kibana will run a [saved objects migration](https://www.elastic.co/guide/en/kibana/current/saved-object-migrations.html) on the default `.kibana` index, or whatever it finds configured as `kibana.index` in `kibana.yml`.
+
+Now, because you may have multiple kibana indices containing saved objects, you should apply the "saved object migration" to those indices as well.
+
+The easiest way to do this is also the most "ignorant" one: 
+
+**Preparation**
+
+* download the newer Kibana version in your laptop (in the example, 8.0.0)
```
Dzuming, line 200: `in your laptop` seems too specific maybe `to your computer` will be better here 🤔 ?

kibana.md
```diff
@@ -181,11 +181,33 @@ $ bin/kibana-plugin remove readonlyrest_kbn
...
+* Restart Kibana
+
+### Major version upgrades when using multi-tenancy
+If you use multi-tenancy (Enterprise only), you will have one or more teanancy-specific kibana indices beyond the main `.kibana` (e.g. `.kibana_tenant1`, `.kibana_tenant2`, etc.).
+
+The first time you run Kibana after a major version upgrade (e.g. upgrading from Kibana 7.17.7 to Kibana 8.0.0), Kibana will run a [saved objects migration](https://www.elastic.co/guide/en/kibana/current/saved-object-migrations.html) on the default `.kibana` index, or whatever it finds configured as `kibana.index` in `kibana.yml`.
+
+Now, because you may have multiple kibana indices containing saved objects, you should apply the "saved object migration" to those indices as well.
+
+The easiest way to do this is also the most "ignorant" one:
```
Dzuming, line 196: Is it any alternative to do it? maybe we should mention it?

kibana.md
```diff
@@ -181,11 +181,33 @@ $ bin/kibana-plugin remove readonlyrest_kbn
...
+
+* download the newer Kibana version in your laptop (in the example, 8.0.0)
+* edit the `kibana.yml` to point to the same Elasticsearch cluster as your production Kibana (`elasticsearch.hosts` setting)
+* add the appropriate `elasticsearch.username` and `elasticsearch.password` credentials settings, same as in production
+
+**For each tenancy index** 
+
+* configure `kibana.yml` with a `kibana.index: <tenancy kibana index>`
+* run Kibana for a few seconds
+* now Kibana will have migrated the tenancy index, like it did with the main `.kibana` index.
```
Dzuming, line 208: Will be any information in logs that, this operation was successfully done? Maybe we should mention it?

kibana.md
```diff
@@ -181,11 +181,33 @@ $ bin/kibana-plugin remove readonlyrest_kbn
...
+
+This is ugly, but the safest way available at the moment.
```
coutoPL, line 210: are we going to provide some tool to migrate it automatically in the future?

kibana.md
```diff
@@ -181,11 +181,33 @@ $ bin/kibana-plugin remove readonlyrest_kbn
...
 * [Unpatch Kibana](./#unpatching-kibana)
-* [Uninstall](./details/kibana-7.8.x-and-older.md#uninstall) the old plugin
-* Delete all the content of "optimize" directory (in the main Kibana installation directory) `rm -rf optimize/`
+* [Uninstall](#uninstalling) the old plugin
 * [Install](./#installation) the new one
 * [Patch Kibana](./#patching-kibana)
-* Restart Kibana.
+* Restart Kibana
+
+### Major version upgrades when using multi-tenancy
```
coutoPL, line 189: who is creating tenant kibana indices? ROR KBN?
If so, it'd be nice to have automatic or semi-automatic migration. If not, maybe it's not strictly related to ROR and this section could be moved to a separate file?

kibana.md
sscarduzio: We could. Probably within `ror-tools.js`.

kibana.md
sscarduzio: Yeah this is a good point. Kibana runs their migration tool at every boot (then eventually returns immediately if nothing is there to be done). We could have that invoked automatically as well, need to find a way to call it though.

Dzuming (approved):

PR #137, opened 2023-01-10
[RORDEV-744] example: self-managed ROR cluster cooperation with remote Elastic Cloud cluster

coutoPL:


coutoPL:
todo: I need to draw the digram (will do it tomorrow)

sscarduzio:
"Here you will be your data" 🤔

coutoPL:
oh boy! 
this happens every time I edit instead of throwing the text out and write it again :P

sscarduzio:
Other than that, this works beautifully and is a pleasure to use. Thank you for the great work! ✨

sscarduzio (approved):

coutoPL:
> curl https://api.beshu.tech/list_es_versions | jq .kbn_universal[0]

are you able to expose also all the ROR version numbers that are available to download without unfreezing?

coutoPL:
@sscarduzio 

I've added to the script (in the ror-sandbox repo):
1. default version of ES, KBN, ROR (latest from the ROR API response)
2. validation of the entered values
3. retrying when the entered value seems to be incorrect 
4. when `jq` is not available on the client's machine, fallback to the old script without default and validations

It would be nice if you could do this: 
https://github.com/beshu-tech/readonlyrest-docs/pull/137#issuecomment-1382326429 

... clients will be able to pick the old version of ROR (maybe it's not necessary, but perhaps we will need it someday)

coutoPL:
@sscarduzio kindly reminder

sscarduzio:
Tested it works, thank you!

sscarduzio:
> > curl https://api.beshu.tech/list_es_versions | jq .kbn_universal[0]
> 
> 
> 
> are you able to expose also all the ROR version numbers that are available to download without unfreezing? 

Not at the moment, could try out something together.

PR #146, opened 2023-01-31
RORDEV-729 Misleading "Empty metadata" message

Dzuming:
- Provide an information about "problem with the configuration of authentication:


kibana.md
```diff
@@ -370,20 +370,44 @@ The solution to this is to reorder the ACL blocks, so the ones that authenticate
...
+
+An example of this would be:
+
+```yaml
+readonlyrest:
+   access_control_rules:
+   - name: "LDAP Auth"
+     ldap_authentication: ...
+   
+   - name: "Allow requests from localhost"
```
sscarduzio, line 399: Let's also propose a solution to the example case: i.e. avoid using hosts rule in general, or at least using it alone without strictly limiting it with some other rules like actions.

Dzuming: Added ✔️

sscarduzio (approved):
Excellent, thank you!

PR #153, opened 2023-03-22
RORDEV-861 provide feature documentation

Dzuming:
- Feature documentation

@sscarduzio I didn't describe this login attempt case, I'm not sure if it's possible in the current implementation, we probably need to find a way to store this kind of information.


kibana.md
```diff
@@ -1101,3 +1101,189 @@ an admin, who has just configured access for a new user. They would like
...
+      });
+    });
+  });
+
+  observer.observe(logoHeader, { childList: true });
+}
+```
+All metadata should be available via `window.ROR_METADATA`. You can use `window.ROR_METADATA.newLogo`. After login as a user with username `admin` you should see a custom logo
+
+#### Reject kibana machine2machine based on a custom metadata
```
sscarduzio, line 1257: Reject machine to machine traffic using custom metadata ACL rules

kibana.md
```diff
@@ -1101,3 +1101,189 @@ an admin, who has just configured access for a new user. They would like
...
+| getIdentitySession()                                                   | IdentitySession or undefined | Get the session identity (check the information below, for the exact response) |
+| setIdentitySession(identitySession: IdentitySession or undefined)      | void                         | Set the new session                                                            |
+| enrichIdentitySessionMetadata(customMetadata: Record<string, unknown>) | void                         | Enrich existing user session by the additional custom metadata                 |
+| lastSessionActivityDate                                                | Date or undefined            | Date of the last session activity.Using in the context of a session timeout    |
+| extractHiddenAppsNames                                                 | string[]                     | List of all hidden apps for specific users |
+
+```ts
+export interface IdentitySession {
+  sid: string;
+  metadata: {
```
coutoPL, line 1139: wait a sec ... do we want to expose all of these? During the analysis, I argued about encapsulation and I was not a fan of exposing ROR internals to the user's code. So we decided to create a metadata field in the `kibana` rule (see [jira](https://readonlyrest.atlassian.net/jira/software/projects/RORDEV/boards/2?selectedIssue=RORDEV-837)). And only this data should be available to the user.

Dzuming: Yeah, that's the plan, and we even expose all metadata for the `kibana_custom_js_inject_file` purpose. So the idea is to expose only custom metadata (I understand the metadata field this way). AFAIR it was a business decision to expose everything to the user WDYT @sscarduzio ?

kibana.md
```diff
@@ -1101,3 +1101,189 @@ an admin, who has just configured access for a new user. They would like
...
+     readonly expiresAt: Date;
+     readonly lastSessionActivityDate?: Date;
+     readonly authorizationHeaders: Map<string, string>;
+     readonly username: string;
+     readonly kibanaHiddenApps: string[];
+     readonly currentGroup?: string;
+     readonly availableGroups: string[];
+     readonly kibanaAccess?: 'ro_strict' | 'ro' | 'rw' | 'admin' | 'unrestricted'
```
sscarduzio, line 1147: this needs to add the new `api_only`

kibana.md
coutoPL: But keeping anything as a public interface is an effort. It's better to have not so many public interfaces. I'm an advocate of exposing as little as possible. Do we know how users would use information about `kibana hidden` apps? or `kibana access`? or impersonation ... or a such internal thing like `correlationId`. 

Currently, I'm working on supporting vars usage in ROR ES `kibana.metadata` value. If we have that why do we need to expose `availableGroups`, `currentGroup`, `username` in the identity object? 

IMO we should expose as little as possible. Narrow the public interface. When we see how users use the feature and know the limitations (from their feedback) we could discuss how to expose more in the safest way. It's very hard to fix public interfaces in a seamless way for the users.

kibana.md
```diff
@@ -1101,3 +1101,189 @@ an admin, who has just configured access for a new user. They would like
...
+| isCookiePresent(cookieName: string)                                    | boolean or undefined         | Check if the specific cookie is presented in the request                       |
+| getIdentitySession()                                                   | IdentitySession or undefined | Get the session identity (check the information below, for the exact response) |
+| setIdentitySession(identitySession: IdentitySession or undefined)      | void                         | Set the new session                                                            |
+| enrichIdentitySessionMetadata(customMetadata: Record<string, unknown>) | void                         | Enrich existing user session by the additional custom metadata                 |
+| lastSessionActivityDate                                                | Date or undefined            | Date of the last session activity.Using in the context of a session timeout    |
+| extractHiddenAppsNames                                                 | string[]                     | List of all hidden apps for specific users |
+
+```ts
```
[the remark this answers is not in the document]

kibana.md
```diff
@@ -1101,3 +1101,189 @@ an admin, who has just configured access for a new user. They would like
...
+        "http_method": "ANY" | Method,
+        "path_regex": string
+     }[];
+  };
+}
+```
+
+### Use cases
+
+**⚠️IMPORTANT** Pre auth custom hook must return `next()` function, to not block the request
```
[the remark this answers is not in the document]

kibana.md
```diff
@@ -1101,3 +1101,189 @@ an admin, who has just configured access for a new user. They would like
...
+
+#### Enriching the metadata
+The metadata is the user-specific data available after the login to the kibana. thanks to the pre-auth custom hook, you can enrich metadata and use them in the kibana custom js file.
```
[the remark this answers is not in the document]
Dzuming: > ...available after the Kibana user successfully logs in. Thanks to...

Done ✔️ 

> This hook is (correctly) called after the auth is processed (successfully or not). So why is it called pre-auth? 🤔
What if we call this feature just custom middleware?

I had the same feeling while writing these examples, I will rename it in our code. Done ✔️

kibana.md

kibana.md
```diff
@@ -1101,3 +1101,187 @@ an admin, who has just configured access for a new user. They would like
...
+| metadata.kibanaAccess                | 'ro_strict' or 'ro' or 'rw' or 'admin' or 'unrestricted' or 'api_only' | 'admin'                                                                      | User kibana access                                                                   |
+| metadata.kibanaIndex                 | string or undefined                                                    | '.kibana_administrators'                                                     | Currently selected group kibana index defined in the ACL                             |
+| metadata.kibanaTemplateIndex         | string or undefined                                                    | '.kibana_template'                                                           | Kibana template index defined in the kibana.yml readonlyrest_kbn.kibanaTemplateIndex |
+| metadata.origin                      | string or undefined                                                    |                                                                              |                                                                                      |
```
Dzuming, line 1282: @sscarduzio @coutoPL  I have a question about the origin which is an `x-ror-origin` what's the function and example of this value ( i didn't find any information in the es/kibana plugin code and docs, and this value is always undefined)

coutoPL: this is how it's described in the swagger:
```
  x-ror-origin:
          type: string
          description: the value taken from JWT's x-ror-origin claim
```

"... when `ror_kbn_auth` is matched" (should be added)

I made this description based on ROR ES code.

Dzuming: Thanks, I forgot that swagger is not a part of the `elasticsearch-readonlyrest-plugin` repository that's why I couldn't find any description :D

kibana.md
```diff
@@ -1101,3 +1101,189 @@ an admin, who has just configured access for a new user. They would like
 to know if the rule(s) are configured correctly. And here comes the impersonation feature. The admin can impersonate given user in Kibana and see what the user would see if they logged in themselves. 
 
 ROR plugins support impersonation and provide UI for configuring cluster before using it. Visit the [impersonation details page](./details/impersonation.md) to know more.
+
+## Custom pre-auth hook
+### Why do we need it
```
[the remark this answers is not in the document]
Dzuming: I added an example and moved the table at the end ✔️

kibana.md
```diff
@@ -1101,3 +1101,189 @@ an admin, who has just configured access for a new user. They would like
...
+    });
+  }
+
+  return next();
+}
+```
+We will pass new logo custom metadata when logged in user username is 'admin'
```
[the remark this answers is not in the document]

kibana.md
```diff
@@ -1101,3 +1101,189 @@ an admin, who has just configured access for a new user. They would like
...
+          handleInit();
+        }
+      });
+    });
+  });
+
+  observer.observe(logoHeader, { childList: true });
+}
+```
+All metadata should be available via `window.ROR_METADATA`. You can use `window.ROR_METADATA.newLogo`. After login as a user with username `admin` you should see a custom logo
```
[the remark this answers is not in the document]
sscarduzio: I referred to the first instance only, "will be available via..." but why not both? :)

kibana.md
sscarduzio: NB: these properties are not exposing the internal messaging format between ES and Kibana, rather they are exposing the configuration parameters of the Kibana+ROR UI. 

* `availableGroups`, `currentGroup`, `username`, `kibana_access`, `hidden_apps` all define the Kibana UI state and front end developers need those available to extend ROR with custom behaviours, or they are going to pick this info out of the DOM, which is REALLY unstable.

* correlation_id is not really internal, is it? it was added for audit IIRC? BTW customMiddleware is the perfect point where to generate custom audit data points, we should expose an API for that (in phase2 of this)

sscarduzio: > If we have that why do we need to expose availableGroups, currentGroup, username in the identity object?

For string composition/manipulation on the ACL side mainly

kibana.md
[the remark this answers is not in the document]
Dzuming: I rephrased it a little, and extend this description

```
All session metadata will be available via `window.ROR_METADATA` property. To get your custom logo, just use `window.ROR_METADATA.newLogo` value. In the example above, after login in as a user with username `admin` you will see a custom logo.
The whole example is a little complex but seems, kibana logo is also a loading indicator, we need to detect the loading state, replace the logo with a spinner, and after the loading, back the custom logo again.
```

kibana.md
sscarduzio: IIRC the x-ror-origin value should take the name of the OIDC/SAML connector, and is used in Kibana to resolve to which "single sign out" URL the browser should be redirected to. 

Imagine you have 2 SAML IdP configured in kibana.yml, the user hits logout, we delete their cookie, and immediately redirect to the logout URL of the IdP. But which one of the two? :)

coutoPL: @Dzuming it's but I've added the description recently and it's in develop (not in master yet)

kibana.md
coutoPL: > > If we have that why do we need to expose availableGroups, currentGroup, username in the identity object?
> 
> For string composition/manipulation on the ACL side mainly

Does it have a real value when you have the same in the context of JS code? You can manipulate the strings with no limitation

sscarduzio: As I said availableGroups, currentGroup, username, kibana_access, hidden_apps all describe the UI state, and they are concepts that are already perceived as "already known to Kibana" in a normal session (i.e. people see the tenancy selector, the username, etc in the UI all the time). 
Writing in the documentation "if you want to use these variables you have to tell Elasticsearch to send them" is a mental shortcircuit.  Plus, I really don't see these basic things changing anytime.

kibana.md
```diff
@@ -1101,3 +1101,201 @@ an admin, who has just configured access for a new user. They would like
 to know if the rule(s) are configured correctly. And here comes the impersonation feature. The admin can impersonate given user in Kibana and see what the user would see if they logged in themselves. 
 
 ROR plugins support impersonation and provide UI for configuring cluster before using it. Visit the [impersonation details page](./details/impersonation.md) to know more.
+
+## Custom middleware
+### Why do we need it
+Sometimes, there is a need to handle non-standard behavior in the plugin.
+We want to make it more flexible and give the option to customize the plugin behavior by the enterprise license users to adjust the product to the business needs.
+There are two options to declare the custom middleware:
+- JS file: `readonlyrest_kbn.custom_middleware_inject_file: '/path/to/your/file.js'`
```
sscarduzio, line 1110: Can we clarify the path can also be relative. And relative to what directory (IIRC it's the same dir where kibana.yml is found)

kibana.md
```diff
@@ -1101,3 +1101,201 @@ an admin, who has just configured access for a new user. They would like
...
+| getPath()                                                              | string                       | '/s/default/internal/bsearch'                                                | Get the request path                                                           |
+| getUrl()                                                               | string                       | '/s/default/internal/bsearch?compress=true'                                  | Get the request URL                                                            |
+| getBody()                                                              | Body or null or undefined    | { id: 42 }                                                                   | Get the request body                                                           |
+| getParams()                                                            | ParamsDictionary             | { id: '1234' }                                                               | Get URL parameters                                                             |
```
sscarduzio, line 1267: Why don't we document only the `rorRequest`methods that are not already evident in the standard Express.js `req` object?

No much value to let them use `getPath`,`getUrl`, `getParams`, `getQueries`, `getHeaders`, `getOriginAddress`, `getMethod`, `getCookies`, ...  when they already have `req`, maybe we can put note about this and add a link to Express.js `req` type definition instead?

kibana.md
```diff
@@ -1101,3 +1101,201 @@ an admin, who has just configured access for a new user. They would like
...
+| metadata.impersonatedBy              | string or undefined                                                    | 'admin'                                                                      | Name of the impersonator                                                             |
+| metadata.correlationId               | string                                                                 | 'd7df607d-4b04-454b-979c-3e962e312a7d'                                       | Current session correlationId                                                        |
+| metadata.customMetadata              | Record<string, unknown>                                                | { metadata1: value }                                                         | Custom metadata defined in the ACL                                                   |
+| metadata.allowedApiPaths.http_method | "ANY" or Method                                                        | 'GET'                                                                        | HTTP method defined as allowed in the ACL                                            |
```
sscarduzio, line 1300: Do they really have the opportunity to modify `allowedApiPaths` at this stage? Or is it already enforced?

kibana.md
```diff
@@ -1101,3 +1101,201 @@ an admin, who has just configured access for a new user. They would like
...
+
+2. Declare custom kibana js file `readonlyrest_kbn.kibana_custom_js_inject_file: '/path/to/custom_kibana.js'`
+
+```js
+async function customMiddleware(req, res, next) {
+   const metadata =
+           req.rorRequest && req.rorRequest.getIdentitySession() && req.rorRequest.getIdentitySession().metadata;
+
+   const headerAuth = req.rorRequest && req.rorRequest.getAuthorizationHeaders && req.rorRequest.getHeaders().getAuthorizationHeaders().get('authorization');
+   const basicHeaderAuth = headerAuth && headerAuth.includes('Basic')
```
sscarduzio, line 1243: this var should be called `isBasicAuth` because it's boolean

sscarduzio (changes requested):
Last comment I have is that all this can be a 3 line paragraph, that mentions a list of possible use cases, and terminates with a link to a new page dedicated to custom middleware

Dzuming:
> Last comment I have is that all this can be a 3 line paragraph, that mentions a list of possible use cases, and terminates with a link to a new page dedicated to custom middleware

I moved everything to the external files and left the link in the kibana.yml file

kibana.md

kibana.md
Dzuming: `allowedApiPaths` will be called before the custom middleware

kibana.md
Dzuming: Right, i removed redundant information and add info about express req and res objects

kibana.md
Dzuming: As I checked it's relative to the kibana root folder, in the case of our local-testing, it's `/kibana` folder. Added the info ✔️

kibana.md
```diff
@@ -1101,3 +1101,201 @@ an admin, who has just configured access for a new user. They would like
 to know if the rule(s) are configured correctly. And here comes the impersonation feature. The admin can impersonate given user in Kibana and see what the user would see if they logged in themselves. 
 
 ROR plugins support impersonation and provide UI for configuring cluster before using it. Visit the [impersonation details page](./details/impersonation.md) to know more.
+
+## Custom middleware
+### Why do we need it
+Sometimes, there is a need to handle non-standard behavior in the plugin.
+We want to make it more flexible and give the option to customize the plugin behavior by the enterprise license users to adjust the product to the business needs.
```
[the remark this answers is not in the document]

sscarduzio (approved):

PR #154, opened 2023-03-26
[RORDEV-236][RORDEV-871] The audit documentation with the new format

mateuszkp96:
The audit documentation with the new format. The deprecated names are still supported, but the example configurations follow the new format


details/kibana-7.8.x-and-older.md
```diff
@@ -811,7 +811,10 @@ Let's start to add to our access control list (found in $ES\_PATH\_CONF/config/r
 
 ```
 readonlyrest:
-  audit_collector: true
+  audit:
+    enabled: true
+    outputs:
```
coutoPL, line 816: `outputs: [index]`

I wonder if we shouldn't assume that `outputs: [index]` is a default setting? (cc @sscarduzio)
So, the minimal configuration could look like this:

```
audit:
  enabled: true
```

or

```
audit.enabled: true
```

We should notice that previous minimal configuration was `audit_collector: true`

elasticsearch.md
```diff
@@ -1451,9 +1451,10 @@ INDEX NOT FOUND req={  ID:501806845-1996085406#74,  TYP:GetIndexRequest,  CGR:N/
 
 The state above is only possible for read-only ES requests \(ES requests which don't change ES cluster state\) for a block containing an `indices` rule. If all other rules within the block are matched, but only the `indices` rule is mismatched, the final state of the block is forbidden due to an index not found.
 
-#### Audit logs
+### Audit
 
-ReadonlyREST can write events very similarly to Logstash into to a series of indices named by default `readonlyrest_audit-YYYY-MM-DD`. Every event contains information about a request and how the system has handled it. Here is an example of the data points contained in each audit event. We can leverage all this information to build interesting Kibana dashboards, or any other visualization.
+ReadonlyREST can write audit events containing information about a request and how the system has handled it.
```
coutoPL, line 1456: where it can write them?

elasticsearch.md
```diff
@@ -1451,9 +1451,10 @@ INDEX NOT FOUND req={  ID:501806845-1996085406#74,  TYP:GetIndexRequest,  CGR:N/
...
+Here is an example of the data points contained in each audit event. We can leverage all this information to build interesting Kibana dashboards, or any other visualization.
 
 ```javascript
```
coutoPL, line 1459: json?

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
   }
 ```
 
-Here is a configuration example, you can see the `audit.collector: true` setting, which normally defaults to false. Note how the successful requests matched by the first rule \(Kibana\) will not be written to the audit log, because the verbosity is set to error. Audit log in facts, obey the verbosity setting the same way regular text logs do.
+#### Configuration
 
-**NB** Following audit configurations are presented with the separate `audit` section, although it can be omitted as settings defined using previous approach (defined directly under `readonlyrest` section and starting with `audit_`)` are still supported (but not recommended) preserving backward compatibility.
+The audit collecting by default is disabled. To enable it, you need to add `audit.enabled: true` and configure the `audit.outputs`.
+With the `outputs` array, you manage the audit sinks configurations.
```
coutoPL, line 1494: `sink` is too a technical word, IMO.  Maybe sth like this:
```
In the `outputs` array, you can define i.a. where the audit events should be sent.
```

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
+You can configure multiple outputs for audit events, but at least one audit output must be defined when the audit is enabled.
```
coutoPL, line 1495: When the audit is enabled, at least one output has to be defined.

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
+The supported output types are:
```
coutoPL, line 1496: The currently supported

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
+* `index` - similarly to Logstash it writes audit events in the documents stored in the indices named by default `readonlyrest_audit-YYYY-MM-DD`
```
coutoPL, line 1497: `in the ReadonlyREST audit index`
(let's write about defaults later)

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
+* `log` - it allows you to collect audit logs in a similar way to the app logs and format them with the help of features that `log4j2` enables.
```
coutoPL, line 1498: 1. the Elasticsearch logs
2. "audit logs" or "audit events"?

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
+  Uses a dedicated logger from the elasticsearch's `log4j2` config.
 
 ```text
```
coutoPL, line 1501: yaml

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
 readonlyrest:
```
coutoPL, line 1502: shouldn't we add info what this example shows?

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
+     type: allow
+     auth_key: kibana:kibana
+     verbosity: error
 
-    - name: "::RO::"
-      auth_key: simone:ro
-      kibana_access: ro
+   - name: "::RO::"
+     auth_key: simone:ro
+     kibana_access: ro
```
coutoPL, line 1518: let's use the new format:
```
  kibana:
    access: ro
```

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
 ```
 
-### Extended audit
+If you want to use the default settings of audit outputs, you can use the simple format of `audit.outputs`:
```
coutoPL, line 1521: IMO we should treat this as a default. The simple version will be probably the most commonly used. The extended version is for superusers.

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
+  audit:
+    enabled: true
+    outputs: 
+    - type: index
+    - type: log
+      enabled: false # by default is true
+    ...
+```
+
+#### The 'index' output specif configurations
```
coutoPL, line 1545: typo: specific

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
+    - type: index
+      cluster: ["https://user1:password@auditNode1:9200", "https://user2:password@auditNode2:9200"]
   ...
 ```
 
-### Custom audit indices name and time granularity
+Setting `audit.cluster` is optional, it accepts non empty list of audit cluster nodes URIs.
 
-It is possible to change the name of the produced audit log indices by specifying a template value as `audit.index_template`.
+#### The 'log' output specif configurations
```
coutoPL, line 1580: typo: specific

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
+    - type: log
+      enabled: false # by default is true
+    ...
+```
+
+#### The 'index' output specif configurations
+
+##### Custom audit indices name and time granularity
+
+It is possible to change the name of the produced audit indices by specifying a template value as `index_template`.
```
coutoPL, line 1549: maybe:
```By default ReadonlyREST audit index name template is `readonlyrest_audit-YYYY-MM-DD`. You can customize the name template using the `index_template` settings. ```

File: elasticsearch.md [hunk 16]
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
 ```
 
-### Custom audit indices name and time granularity
+Setting `audit.cluster` is optional, it accepts non empty list of audit cluster nodes URIs.
 
-It is possible to change the name of the produced audit log indices by specifying a template value as `audit.index_template`.
+#### The 'log' output specif configurations
 
-Example: tell ROR to write on monthly index.
+The `log` output uses a dedicated logger to write the audit events in files. The logger applies the `INFO` log level to the appended audit events.
```
coutoPL, line 1582: ```events to the Elasticsearch log at INFO level```
the second sentence can be deleted

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
 
+To configure audit events writing to logs, you need to add to the `readonlyrest` configuration the following settings:
```
coutoPL, line 1584: ```To make ReadonlyREST start adding the audit events to the Elasticsearch log, all you have to do is add "log" as one of the outputs, e.g:```

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
-    collector: true
-    index_template: "'custom-prefix'-yyyy-MM"  # <--monthly pattern
+    enabled: true
+    outputs: 
+    - type: log
   ...
 ```
 
-**⚠️IMPORTANT**: notice the single quotes inside the double quoted expression. This is the same syntax used for [Java's SimpleDateFormat](https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html).
+In the next step you need to configure the logger for audit events in the `$ES_PATH_CONF/config/log4j2.properties`. Here is an example config with default logger name:
```
coutoPL, line 1594: Is it needed? I thought it's an optional part (when eg. you want to have a separate file with audit logs)

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
+
+logger.readonlyrest_audit.name = readonlyrest_audit   # required logger name, must be the same as the one defined in `readonlyrest.yml` 
+logger.readonlyrest_audit.appenderRef.ror_audit.ref = readonlyrest_audit_rolling    
+logger.readonlyrest_audit.additivity = false          # set to false to use only desired appenders
+```
+All settings are up to you. The only required entry is the logger name `logger.{your-logger-name}.name = {your-logger-name}`.
+The default logger name is the `readonlyrest_audit`.
+
+If you want to set a custom logger name for the `log` output, add the `logger_name` setting for the given output:
+```text
```
coutoPL, line 1617: yaml

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
+Setting `audit.cluster` is optional, it accepts non empty list of audit cluster nodes URIs.
 
-It is possible to change the name of the produced audit log indices by specifying a template value as `audit.index_template`.
+#### The 'log' output specif configurations
 
-Example: tell ROR to write on monthly index.
+The `log` output uses a dedicated logger to write the audit events in files. The logger applies the `INFO` log level to the appended audit events.
 
+To configure audit events writing to logs, you need to add to the `readonlyrest` configuration the following settings:
 ```text
```
coutoPL, line 1585: yaml

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
+  ...
+```
+**⚠️IMPORTANT**: notice the single quotes inside the double quoted expression. This is the same syntax used for [Java's SimpleDateFormat](https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html).
+
+##### Custom audit cluster
 
-If you want to log the request content then an additional serializer is provided. This will log the entire user request within the content field of the audit event. To enable, configure the audit\_serializer parameter as below.
+It's possible to set a custom audit cluster responsible for audit events storage. When a custom cluster is specified, items will be sent to defined cluster nodes instead of the local one.
 
 ```text
```
coutoPL, line 1568: yaml

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
+
+#### The 'index' output specif configurations
+
+##### Custom audit indices name and time granularity
+
+It is possible to change the name of the produced audit indices by specifying a template value as `index_template`.
+
+Example: tell ROR to write on the monthly index.
+
+```text
```
coutoPL, line 1553: yaml

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
+
+  audit:
+    enabled: true
+    outputs: [ index, log ]
+
+    ...
+```
+
+In case when you want to disable given output, add the `enabled: false` to the output config:
+```text
```
coutoPL, line 1534: yaml

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
-      kibana_access: ro
+   - name: "::RO::"
+     auth_key: simone:ro
+     kibana_access: ro
 ```
 
-### Extended audit
+If you want to use the default settings of audit outputs, you can use the simple format of `audit.outputs`:
+
+```text
```
coutoPL, line 1523: yaml

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
+  audit:
+    enabled: true
+    outputs: 
+    - type: log
+      logger_name: custom-logger-name
+  ...
+```
 
-### Custom audit log serializer
+### Extended audit
```
coutoPL, line 1627: Extending audit events

elasticsearch.md
```diff
@@ -1487,63 +1488,166 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
...
 
-You can write your own custom audit log serializer class, add it to the ROR plugin class path and configure it through the YAML settings.
+If you want to log the request content then an additional serializer is provided. This will log the entire user request within the content field of the audit event. To enable, configure the `serializer` parameter as below.
```
coutoPL, line 1629: IMO we should link the audit event example starting in line 1459 and say that this is the default format of our audit event. And mention that it can be customized using sth we call Serializer. We should also add info how the default serializer is called (it can be extended by our users). Then we should about other predefined serializers.

coutoPL (changes requested):

details/kibana-7.8.x-and-older.md
sscarduzio: empty or absent outputs array should imply outputs=index, and +1 for the backward compat.

coutoPL: AFAIR the empty array is an invalid state and IMO it's better to leave it like this.

sscarduzio: Yeah makes sense!

elasticsearch.md
mateuszkp96: Yes, it's optional. I think I was focused too much on custom logging. 
This is how the log looks like without additional changes in log4j properties:
```[2023-04-04T20:14:34,415][INFO ][readonlyrest_audit       ] {"error_message":"error","headers":[],"acl_history":"","origin":"localhost","match":false,"final_state":"ERRORED","destination":"localhost","task_id":0,"type":"default-type","req_method":"GET","path":"_search","indices":[],"@timestamp":"2018-12-31T23:01:59Z","content_len_kb":0,"error_type":"Exception","correlation_id":"7029bd54-2e0c-4e10-a137-933f15efb52e","processingMillis":134334755414,"action":"default-action","block":"error","id":"6ff2b948-d509-4bf1-be3f-4bcda0b4f0fb","content_len":0}```

Comment (coutoPL) [reply elasticsearch.md:1594]: It's ok, but we should clearly mention that I user wants to have the audit in separate file, they can do it like that

File: details/kibana-7.8.x-and-older.md (hunk 1 above)
Comment (mateuszkp96) [reply details/kibana-7.8.x-and-older.md:816]: Fixed

File: elasticsearch.md [hunk 27]
```diff
@@ -1501,63 +1501,178 @@ ReadonlyREST can write events very similarly to Logstash into to a series of ind
   }
 ```
 
-Here is a configuration example, you can see the `audit.collector: true` setting, which normally defaults to false. Note how the successful requests matched by the first rule \(Kibana\) will not be written to the audit log, because the verbosity is set to error. Audit log in facts, obey the verbosity setting the same way regular text logs do.
+#### Configuration
```
coutoPL, line 1504: I want to ask you about one more refactor. The configuration part of the auditing feature is huge. Let's move it to the `details` folder to a separate file and link it here.

Please, don't forget by modifying SUMMARY.md (for a gitbook's sake)

coutoPL (changes requested):

coutoPL (approved):
LGTM. @sscarduzio could you please take a look too?

details/audit.md
```diff
@@ -0,0 +1,270 @@
...
+{
+    "error_message": null,
+    "headers": [
+      "Accept",
+      "Authorization",
+      "content-length",
+      "Host",
+      "User-Agent"
+    ],
+    "acl_history": "[[::LOGSTASH::->[auth_key->false]], [::RW::->[kibana_access->true, indices->true, kibana_hide_apps->true, auth_key->true]], [kibana->[auth_key->false]], [::RO::->[auth_key->false]]]",
```
sscarduzio, line 15: kibana_access => access ? (etc)

coutoPL: it's fixed

PR #160, opened 2023-05-22
RORDEV-898_Provide_automatic_migration_of_indexes_when_using_multi-tenancy

Dzuming:
- Provide automatic migration_of indexes when using multi-tenancy info

### DIFF
--- a/kibana.md
+++ b/kibana.md
@@ -167,18 +167,9 @@ The first time you run Kibana after a major version upgrade (e.g. upgrading from
 
 Now, because you may have multiple kibana indices containing saved objects, you should apply the "saved object migration" to those indices as well.
 
-The only known safe - yet pedantic - way to do this is the following process: 
+All tenancy kibana index are migrated automatically.
 
-**Preparation**
-
-* download the newer Kibana version in your computer (in the example, 8.0.0)
-* edit the `kibana.yml` to point to the same Elasticsearch cluster as your production Kibana (`elasticsearch.hosts` setting)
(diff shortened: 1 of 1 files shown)


kibana.md
```diff
@@ -167,18 +167,9 @@ The first time you run Kibana after a major version upgrade (e.g. upgrading from
...
-
-* download the newer Kibana version in your computer (in the example, 8.0.0)
-* edit the `kibana.yml` to point to the same Elasticsearch cluster as your production Kibana (`elasticsearch.hosts` setting)
-* add the appropriate `elasticsearch.username` and `elasticsearch.password` credentials settings, same as in production
-
-**For each tenancy index** 
-
-* configure `kibana.yml` with a `kibana.index: <tenancy kibana index>`
-* run Kibana for a few seconds, until you see something similar to these log lines:
+* After opening a specific tenancy kibana index for example after the tenancy switch or login to the kibana there will be these logs visible:
```
sscarduzio, line 172: let's explain it a bit more throughly:

---

ReadonlyREST Enterprise will automatically make sure a tenancy index is migrated to satisfy the current Kibana version  **right before every time it's being used**. 

For example, after a tenant logs in, before the Kibana session is started, or when a user changes tenancy with the tenancy switcher, the tenancy index gets created if absent, checked and migrated if necessary.

Dzuming: Updated ✔️

sscarduzio (approved):

PR #165, opened 2023-06-27
RORDEV-921 Provide a way to set the default tenant during the login to the Kibana and change the order of tenants in the ROR menu by the “Custom Middleware”

Dzuming:
- Provide a way to set the default tenant during the login to the Kibana and change the order of tenants in the ROR menu by the “Custom Middleware”

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -32,6 +32,7 @@
   * [Custom middleware](examples/custom-middleware/README.md)
     * [Enriching the metadata](examples/custom-middleware/enriching-the-metadata.md)
     * [Reject machine-to-machine traffic using custom metadata ACL rules](examples/custom-middleware/reject-machine-to-machine-traffic-using-custom-metadata-ACL-rules.md)
+    * [Tenant group manipulation](examples/custom-middleware/tenant-group-manipulation.md)
     * [Available rorRequest API](examples/custom-middleware/available-ror-request-api.md)
     * [User Session identity](examples/custom-middleware/reject-machine-to-machine-traffic-using-custom-metadata-ACL-rules.md)
 * [Contribution License Agreement](contributing.md)
--- a/examples/custom-middleware/tenant-group-manipulation.md
+++ b/examples/custom-middleware/tenant-group-manipulation.md
@@ -1,0 +1,46 @@
+---
+description: Tenant group manipulation
+---
+
+# Tenant group manipulation
+
+We can change a default tenant and order of the tenants in the ROR menu by providing `defaultGroup` query string parameter before the login, and chenge the order of `availableGroups`
+thanks to the `enrichIdentitySessionMetadata` method.
+
+1. Declare `readonlyrest_kbn.custom_middleware_inject_file: 'path/to/custom_middleware_inject_file.js'` in the kibana.yml and declare `custom_middleware_inject_file.js`
+
+```js
+async function customMiddleware(req, res, next) {
+    const rorRequest = req.rorRequest;
+    const metadata =
+        req.rorRequest && req.rorRequest.getIdentitySession() && req.rorRequest.getIdentitySession().metadata;
+    const defaultGroup = 'infosec';
+
+    if (rorRequest.getPath() === '/login' && rorRequest.getMethod() === 'post') {
+        if (rorRequest.getBody().username === 'admin') {
+            rorRequest.setQuery('defaultGroup', defaultGroup);
+        }
+    }
+
+    if (metadata && rorRequest.getPath() === '/pkp/api/info') {
+        const availableGroups = metadata.availableGroups;
+        if (availableGroups.some(availableGroup => availableGroup === defaultGroup)) {
+            const index = availableGroups.indexOf(defaultGroup);
+            const groupAvailable = index !== -1;
+            if (groupAvailable) {
+                availableGroups.splice(index, 1);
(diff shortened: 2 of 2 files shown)


SUMMARY.md
```diff
@@ -32,6 +32,7 @@
   * [Custom middleware](examples/custom-middleware/README.md)
     * [Enriching the metadata](examples/custom-middleware/enriching-the-metadata.md)
     * [Reject machine-to-machine traffic using custom metadata ACL rules](examples/custom-middleware/reject-machine-to-machine-traffic-using-custom-metadata-ACL-rules.md)
+    * [Tenant group manipulation](examples/custom-middleware/tenant-group-manipulation.md)
```
sscarduzio, line 35: I'd call this task "Reordering available tenancies"

examples/custom-middleware/tenant-group-manipulation.md
```diff
@@ -0,0 +1,46 @@
+---
+description: Tenant group manipulation
+---
+
+# Tenant group manipulation
+
+We can change a default tenant and order of the tenants in the ROR menu by providing `defaultGroup` query string parameter before the login, and chenge the order of `availableGroups`
```
sscarduzio, line 7: "We can change the default tenancy, and the display ordering of the tenancies in the ROR menu by providing the `defaultGroup` query parameter in the HTTP request submitted by the login form."

examples/custom-middleware/tenant-group-manipulation.md
```diff
@@ -0,0 +1,46 @@
...
+            rorRequest.enrichIdentitySessionMetadata({ availableGroups });
+        }
+    }
+
+    return next();
+}
+```
+
+In this example, before the login to the Kibana, when the username is equal 'admin', we add default tenant `rorRequest.setQuery('defaultGroup', defaultGroup);` which means,
+that it will be the first tenant opened after the login. After the login, we will change the order of tenants displayed in the ROR menu and our default tenant will be the first on the list.
```
sscarduzio, line 44: Avoid repetition:

"...after the login. During the active Kibana session, we will also change..."

examples/custom-middleware/tenant-group-manipulation.md
```diff
@@ -0,0 +1,46 @@
...
+
+**⚠️IMPORTANT** Custom middleware must return `next()` function, to not block the request
```
sscarduzio, line 46: We should mention the limitations of this approach. This will not work when the ROR login form is not directly involved in the login process. This happens when users log in through SAML/OIDC or reverse proxy (`proxy_pass` rule).

Right?

sscarduzio (approved):

examples/custom-middleware/tenant-group-manipulation.md
sscarduzio: @Dzuming when this is merged, let's add a deep link from the changelog entry to this page (using the docs.readonlyrest.com URL rather than direct github of course)

Dzuming: > We should mention the limitations of this approach. This will not work when the ROR login form is not directly involved in the login process. This happens when users log in through SAML/OIDC or reverse proxy (proxy_pass rule).

I'm not sure if i understand, what kind of problem can you see here? IMO No matter which, method to login the user will use, it will be always `POST /login` call which will be intercepted by the custom middleware. The only problem here, is that in case of SAML/OIDC to receive username, user needs to decrypt JWT token, but i can add it to the examples.

examples/custom-middleware/tenant-group-manipulation.md
Dzuming: Done ✔️

examples/custom-middleware/tenant-group-manipulation.md
Dzuming: Done ✔️

SUMMARY.md
Dzuming: Done ✔️

examples/custom-middleware/tenant-group-manipulation.md
sscarduzio: Hmm what about the proxy pass? IIRC in ROR there's a middleware that catches X-Forwarded-User on the fly without requiring a POST /login?

Dzuming: Good point, but in theory user could provide a default group by something like

```
  const metadata =
    req.rorRequest && req.rorRequest.getIdentitySession() && req.rorRequest.getIdentitySession().metadata;
  const defaultGroup = 'infosec';
 const X_FORWARDED_USER = 'x-forwarded-user';

  if (!metadata && rorRequest.getHeaders()[X_FORWARDED_USER]) {
    if (rorRequest.getHeaders()[X_FORWARDED_USER] === 'admin') {
      rorRequest.setQuery('defaultGroup', defaultGroup);
    }
```
Unfortunately, handleProxyAuth middleware is running before a custom middleware so it won't work, I will create a Jira for it and fix it. To sum up:

- I will provide a SAML/OIDC example
- I will add info that proxy auth not support it yet
- I will create a Jira to call custom middleware before proxy auth

Dzuming: Updated ✔️

sscarduzio (approved):

PR #172, opened 2023-07-29
Change upload file example and structure

Dzuming:
- Change the upload file example
- Change the structure of the request to match the endpoint request body, to be an object with a file field


sscarduzio:
what's the rationale for removing it? It's not being used?

Dzuming:
> what's the rationale for removing it? It's not being used?

I updated the description because it's not the truth anymore. The upload file example was a copy of update settings endpoint, which is not the truth

coutoPL:
@sscarduzio ping

sscarduzio:
I don't understand, the old post json won't work anymore?

coutoPL:
AFAIU we have two endpoints to update ROR settings in the index:
1. using ROR yaml in the request body
2. using file upload

The PR contains the fix of the example and the fixed description of the body of the 2nd option in Open API

sscarduzio:
Oh shoot. Ok. So we are not dropping any historical api?

coutoPL:
I don't know what you mean by historical. The API exposed by ES ROR is slightly different. The one is much more clean than the internal one

PR #177, opened 2023-08-27
[RORDEV-967] Variables functions documentation

mateuszkp96:


elasticsearch.md
```diff
@@ -1908,8 +1908,130 @@ indices: ["logstash_@explode{x-indices_csv_string}*", "otherIdx"]
 # -> indices: ["logstash_a*", "logstash_b*", "otherIdx"]
 ```
 
+### Variables functions
+
+A value resolved from a variable may not be valid in some contexts.
+Sometimes, the value from the variable needs some preprocessing before usage.
+For example, a HTTP header value containing the uppercase characters is a wrong candidate for the index name because it has to be a lowercase string.
+We introduced variable functions to overcome these limitations.
+They allow modification of the variable values during the variable resolution ([static](elasticsearch.md#static-variables) and [dynamic](elasticsearch.md#dynamic-variables) variables are supported).
```
coutoPL, line 1917: `(both, [static](...) and [dynamic](...)`

elasticsearch.md
```diff
@@ -1908,8 +1908,130 @@ indices: ["logstash_@explode{x-indices_csv_string}*", "otherIdx"]
...
+
+With their help, you can use the HTTP header `x-app: KIBANA` containing uppercase characters in the `indices` rule:
+
+```yaml
+indices: [ 'index_@{header:x-app}#{to_lowercase}' ]
```
coutoPL, line 1922: let's explain what it means (eg. in the comment)

elasticsearch.md
```diff
@@ -1908,8 +1908,130 @@ indices: ["logstash_@explode{x-indices_csv_string}*", "otherIdx"]
...
 
+### Variables functions
+
+A value resolved from a variable may not be valid in some contexts.
+Sometimes, the value from the variable needs some preprocessing before usage.
+For example, a HTTP header value containing the uppercase characters is a wrong candidate for the index name because it has to be a lowercase string.
+We introduced variable functions to overcome these limitations.
+They allow modification of the variable values during the variable resolution ([static](elasticsearch.md#static-variables) and [dynamic](elasticsearch.md#dynamic-variables) variables are supported).
+
+With their help, you can use the HTTP header `x-app: KIBANA` containing uppercase characters in the `indices` rule:
```
coutoPL, line 1919: it's rather an unrealistic example. WDYT about using the `X-Forwarded-User` header as the example instead?

elasticsearch.md
```diff
@@ -1908,8 +1908,130 @@ indices: ["logstash_@explode{x-indices_csv_string}*", "otherIdx"]
...
+```
+
+#### Syntax
+
+In general, functions syntax is as follows:
+
+```function_name("arg1","arg2")```
+* `function_name` - a function that you want to apply
+* `(...)` - function call parentheses (they may be omitted when the function has no args)
+* `arg1`, `arg2` - arguments passed to function. They should be surrounded by `"`. If your argument contains a `"` you can escape it with a `\`.
```
coutoPL, line 1938: please add an example which shows the escaping

elasticsearch.md
```diff
@@ -1908,8 +1908,130 @@ indices: ["logstash_@explode{x-indices_csv_string}*", "otherIdx"]
...
+
+You can chain functions with the `.` operator (functions are applied in order from left to right):
+
+```function_a("arg1").function_b.function_c("arg1")```
+
+To apply functions to the variable, you need to use the `#` operator and enter your code in `{ }` braces:
+```text
+@{--variable-definition--}#{--functions-code--}
```
coutoPL, line 1946: `--functions-chain--`?

elasticsearch.md
```diff
@@ -1908,8 +1908,130 @@ indices: ["logstash_@explode{x-indices_csv_string}*", "otherIdx"]
...
+```yaml
+# Using JWT claims as dynamic variables with variable function
+indices: [ ".kibana_@{jwt:department}#{to_lowercase}", "otherIdx" ]
+# claims = { "user": "u1", "department": "Infosec"}
+# -> indices: [".kibana_infosec", "otherIdx"]
+```
+
+#### Supported functions
+
+The list of functions is limited. These are the supported functions:
```
coutoPL, line 1957: maybe it'd be better to say that currently, we support functions like this ... but it could be easily extended. And let's leave a hint that users can let us know what they need and we consider to add the proper implementation

sscarduzio (approved):
Good catch with all those typos. The new  functions system look useful.

coutoPL (changes requested):

coutoPL (approved):
LGTM 👍

PR #179, opened 2023-09-10
RORDEV 976 Provide info that cookiePass is a required kibana.yml property

Dzuming:
- Provide info that cookiePass is a required kibana.yml property from version 1.51.0

### DIFF
--- a/kibana.md
+++ b/kibana.md
@@ -594,6 +594,9 @@ xpack.watcher.enabled: false
 # Kibana server use ::KIBANA-SRV:: credentials
 elasticsearch.username: "kibana"
 elasticsearch.password: "kibana"
+
+# ReadonlyREST required properties
+readonlyrest_kbn.cookiePass: '12312313123213123213123abcdefghijklm'
 ```
 
 And of course also make sure `elasticsearch.url` points to the designated Elasticsearch instance (check also the http or https)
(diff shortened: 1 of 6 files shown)


kibana.md
```diff
@@ -965,6 +965,11 @@ Each Kibana node stores user sessions in-memory. This will cause problems when u
    * `readonlyrest_kbn.sessions_probe_interval_seconds: 120` (default 60s) how often should the browser poll Kibana to check if their session is still valid. Raise this value if you connect to Kibana through slow networks (i.e. VPN), or have very slow loading dashboards.
 3. Add the above config in all Kibana nodes behind the load balancer, and restart them.
 
+
+{% hint style="warning" %}
+From readonlyREST version 1.51.0 `readonlyrest_kbn.cookiePass` is a required `kibana.yml` config parameter.
```
coutoPL, line 970: IMO there is a problem with the fact that we describe cookiePass in case of multiple kbn instances. But now it's always required. Even in ROR Free case.

Dzuming: What's the suggestion? Additional section with a description of why we need this change in the context of the activation key?  IMO  Since the cookiePass is now required, no matter way, we should leave this message here

coutoPL: I'd suggest adding the info somewhere before the section related to configure several kbn instances. The thing you added so far is ok. I'm not telling you should get rid of it.

And it would be nice to review all md files to check if there is no other references to minimal kibana.yml file.

Dzuming: > And it would be nice to review all md files to check if there are no other references to the minimal kibana.yml file.

And also change our ror-sandbox kibana.yml config

coutoPL:
It ok for me

PR #183, opened 2023-10-22
Rordev 966 provide activate license endpoint to the ReadonlyREST API

Dzuming:
- provide activate license endpoint to the ReadonlyREST API

### DIFF
--- a/.gitbook/assets/ReadonlyREST-api.yml
+++ b/.gitbook/assets/ReadonlyREST-api.yml
@@ -91,7 +91,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/badRequestResponse'
+                $ref: '#/components/schemas/BadRequestResponse'
               examples:
                 YamlFileExtensionRequiredResponseExample:
                   $ref: "#/components/examples/YamlFileExtensionRequiredResponseExample"
@@ -141,7 +141,7 @@ paths:
         - $ref: '#/components/parameters/XRorCorrelationIdHeader'
         - in: query
           name: overwrite
-          description: You can set value as true to overwrite existing license key
+          description: You can set value as true to overwrite existing license key. Default value is false.
           schema:
             type: boolean
             example: true
@@ -150,14 +150,14 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/updateLicenseKey'
+              $ref: '#/components/schemas/UpdateLicenseKey'
       responses:
         200:
           description: Returns a status of updating License key
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/updateLicenseKey'
+                $ref: '#/components/schemas/UpdateLicenseKey'
               examples:
                 UpdateLicenseKeySuccessResponse:
                   $ref: "#/components/examples/UpdateLicenseKeySuccessResponse"
@@ -168,7 +168,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/badRequestResponse'
+                $ref: '#/components/schemas/BadRequestResponse'
               examples:
                 JwtTokenExpired:
                   $ref: '#/components/examples/UpdateLicenseKeyExpiredResponse'
@@ -208,7 +208,7 @@ components:
       scheme: basic
 
   schemas:
-    badRequestResponse:
+    BadRequestResponse:
       type: object
       required:
         - message
@@ -304,7 +304,7 @@ components:
           type: string
           description: Human readable message
 
-    updateLicenseKey:
+    UpdateLicenseKey:
       type: object
       description: License key to update
       required:
--- a/.gitbook/assets/ReadonlyREST-api.yml
+++ b/.gitbook/assets/ReadonlyREST-api.yml
@@ -170,14 +170,10 @@ paths:
               schema:
                 $ref: '#/components/schemas/BadRequestResponse'
               examples:
-                JwtTokenExpired:
+                LicenseKeyExpired:
                   $ref: '#/components/examples/UpdateLicenseKeyExpiredResponse'
-                JwtTokenMalformed:
+                licenseKeyMalformed:
                   $ref: '#/components/examples/UpdateLicenseKeyMalformedResponse'
-                JwtTokenInvalid:
-                  $ref: "#/components/examples/UpdateLicenseKeyInvalidResponse"
-                JwtInvalidSignature:
-                  $ref: "#/components/examples/UpdateLicenseKeyInvalidSignatureResponse"
                 NoLicenseKeyProvided:
                   $ref: "#/components/examples/UpdateLicenseKeyNoLicenseKeyResponse"
         403:
@@ -312,9 +308,8 @@ components:
       properties:
         license:
           type: string
-          description: JWT token
+          description: License key
           example: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c
-
 
   parameters:
     XRorCorrelationIdHeader:
@@ -399,23 +394,13 @@ components:
     UpdateLicenseKeyExpiredResponse:
       summary: License key expired
       value:
-        status: JWT_EXPIRED
-        message: "The JWT token provided has expired."
+        status: LICENSE_KEY_EXPIRED
+        message: "The license key provided has expired."
     UpdateLicenseKeyMalformedResponse:
       summary: License key malformed
       value:
-        status: JWT_MALFORMED
-        message: "JWT malformed. The token does not have three components (delimited by a .)."
-    UpdateLicenseKeyInvalidResponse:
-      summary: License key invalid
-      value:
-        status: JWT_INVALID
-        message: "Invalid JWT. The header or payload could not be parsed."
-    UpdateLicenseKeyInvalidSignatureResponse:
-      summary: License key invalid signature
-      value:
-        status: INVALID_JWT_SIGNATURE
-        message: "The provided JWT token has an incorrect or tampered signature, indicating a potential unauthorized modification of the token contents."
+        status: LICENSE_KEY_MALFORMED
+        message: "License key malformed. The token does not have three components (delimited by a .)."
     UpdateLicenseKeyNoLicenseKeyResponse:
       summary: License key not provided
       value:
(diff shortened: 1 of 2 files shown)


.gitbook/assets/ReadonlyREST-api.yml
```diff
@@ -131,6 +131,73 @@ paths:
...
+                NoLicenseKeyProvided:
+                  $ref: "#/components/examples/UpdateLicenseKeyNoLicenseKeyResponse"
+        403:
+          description: The request was forbidden by ReadonlyREST
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ForbiddenRorResponse'
+              examples:
+                notSufficientPermissions:
```
sscarduzio, line 190: InsufficientPermissions?

universal-builds/universal-builds.md
```diff
@@ -36,7 +36,7 @@ $ bin/kibana
 Once Kibana is up and running, you can send a HTTP request to Kibana.
 
 ```
-POST http://<kibana-host-with-ror>:5601/pkp/api/license?overwrite=true
+POST http://<kibana-host-with-ror>:5601/api/ror/license?overwrite=true
```
sscarduzio, line 39: Let's remember to write this change in the release notes

.gitbook/assets/ReadonlyREST-api.yml
Dzuming: Done ✔️

.gitbook/assets/ReadonlyREST-api.yml
```diff
@@ -131,6 +131,73 @@ paths:
...
+    post:
+      summary: Upgrading the license to PRO or Enterprise edition
+      description: The endpoint can be use to Upgrading the license to PRO or Enterprise edition.
+      tags:
+        - License
+      parameters:
+        - $ref: '#/components/parameters/XRorCorrelationIdHeader'
+        - in: query
+          name: overwrite
+          description: You can set value as true to overwrite existing license key
```
coutoPL, line 144: is it a required query param?

.gitbook/assets/ReadonlyREST-api.yml
```diff
@@ -131,6 +131,73 @@ paths:
...
+          schema:
+            type: boolean
+            example: true
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/updateLicenseKey'
```
coutoPL, line 153: maybe `UpdateLicenseKey` to be consistent with other schemas naming

.gitbook/assets/ReadonlyREST-api.yml
```diff
@@ -131,6 +131,73 @@ paths:
...
+      responses:
+        200:
+          description: Returns a status of updating License key
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/updateLicenseKey'
```
coutoPL, line 160: the same here

.gitbook/assets/ReadonlyREST-api.yml
```diff
@@ -131,6 +131,73 @@ paths:
...
+                UpdateLicenseKeySuccessResponse:
+                  $ref: "#/components/examples/UpdateLicenseKeySuccessResponse"
+                UpdateLicenseKeyNotModifiedResponse:
+                  $ref: "#/components/examples/UpdateLicenseKeyUnmodifiedResponse"
+        400:
+          description: Bad Request
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/badRequestResponse'
```
coutoPL, line 171: the same here

.gitbook/assets/ReadonlyREST-api.yml
```diff
@@ -307,6 +386,46 @@ components:
...
+        message: "ok"
+    UpdateLicenseKeyUnmodifiedResponse:
+      summary: License key not modified
+      value:
+        status: UNMODIFIED
+        message: "License token already exists is not modified. If you want to overwrite your license key use overwrite=true query string parameter."
+    UpdateLicenseKeyExpiredResponse:
+      summary: License key expired
+      value:
+        status: JWT_EXPIRED
```
coutoPL, line 402: does it matter to the user that the License is JWT? I think JWT is an implementation detail and we should not bother users with this info

.gitbook/assets/ReadonlyREST-api.yml
```diff
@@ -307,6 +386,46 @@ components:
...
+        message: "The JWT token provided has expired."
+    UpdateLicenseKeyMalformedResponse:
+      summary: License key malformed
+      value:
+        status: JWT_MALFORMED
```
coutoPL, line 407: Continuing the thread ... if the answer is "yes, JWT is the implementation detail", should we explain in the statuses what's wrong with the JWT? We can write in the error log what is exactly wrong, but in API it's better to be generic, if user cannot (or should not) understand what we mean.

.gitbook/assets/ReadonlyREST-api.yml
Dzuming: > does it matter to the user that the License is JWT?

The user who will configure the license key will be the Admin guy who knows what's the JWT (that's my assumption). So in theory it gives additional information to the end user. I have a feeling that it's the same type of implementation detail as the cookie, which on the one side it's implementation details but on the other, it gives the user, useful information.

> I think JWT is an implementation detail and we should not bother users with this info

The problem I can see in the implementation details is that it's something that can change in the future because it was a developer's decision to implement it this way, but is it really a problem in our case?

.gitbook/assets/ReadonlyREST-api.yml
Dzuming: > if the answer is "yes, JWT is the implementation detail", should we explain in the statuses what's wrong with the JWT?

I don't understand this sentence. Do you mean "If we established that it's okay to use JWT token information in our external API, let's not try to describe in the status what's going on"?


> should we explain in the statuses what's wrong with the JWT?

Okay, so the proposition is to use BAD_REQUEST for all 400 errors and describe in the message, what's the exact problem (what we already did). 

> but in API it's better to be generic

I don't have a lot of experience with exposing public API, so my question is maybe a little stupid, but why it's better to be generic? I assume that if someone, wants to use our external API in the "automated way", he/she will want to react to the specific situation automatically, based on the status. So In our situation, if the `JWT_EXPIRED` token, status is, let's send an email (or something like this. :P)

> if user cannot (or should not) understand what we mean.

The question is, is it true in our case?  I mean, IMO since it's a technical user he/she will know what's going on here, and I don't see any problems with exposing this kind of information to the end user, day you have on mind specific case?

.gitbook/assets/ReadonlyREST-api.yml
coutoPL: At the moment the license is JWT. I've just checked my license and what I see is a string (because it's JWT encoded using base64). So, from the perspective of our user, we give him some string that we call the license (through Portal), and they can upload it using ROR KBN (using API or through the UI). 

The fact that he deals with JWT should not bother them. For them, it's just a string (at least knowing that it's a JWT doesn't bring anything important for the user ... that's why, it's better to not bother them).

Then if it's really the implementation detail which is not important for users, we should not mislead them by introducing sth in the API that they could not understand.

.gitbook/assets/ReadonlyREST-api.yml
coutoPL: > I don't understand this sentence. Do you mean "If we established that it's okay to use JWT token information in our external API, let's not try to describe in the status what's going on"?

I mean that using status `JWT_MALFORMED` we say sth like this: "hey, the JWT you uploaded is malformed", but IMO we should say: "hey, the license you uploaded is malformed". So, IMO `MALFORMED_LICENSE` would be better.

> Okay, so the proposition is to use BAD_REQUEST for all 400 errors and describe in the message, what's the exact problem (what we already did).

Yes, using the message is better, because the message text cannot be treated as a part of the API. The message can easily be changed in the future. But the message should follow the same rules the API should follow: we should not mention about things that users are not aware of or about sth they are able to understand (after research) but what doesn't bring anything useful to the context.

> I don't have a lot of experience with exposing public API, so my question is maybe a little stupid, but why it's better to be generic? I assume that if someone, wants to use our external API in the "automated way", he/she will want to react to the specific situation automatically, based on the status. So In our situation, if the JWT_EXPIRED token, status is, let's send an email (or something like this. :P)

No, the question is a good one. It's all about maintenance and backward compatibility of the API. Let's assume that in the future we decide to hide information passed in the JWT token and we decide to encrypt it. It's not a JWT anymore. But it's still a license. But in the API we have JWT all over the place :) What we can do? We can break the compatibility and remove the "JWT" abbrev from the API or not break it and decide to live with sth which is not true anymore. 

During the designing of the API, we were focused too much on the current implementation, but always we should focus on the business side of the API and take into consideration the knowledge the user has (or needs to have).

> The question is, is it true in our case? I mean, IMO since it's a technical user he/she will know what's going on here, and I don't see any problems with exposing this kind of information to the end user, day you have on mind specific case?

Let's determine how users work with the ROR license (correct me when I'm wrong). 
1. they go to the Customer Portal and generate a license
2. they copy the license (the long string)
3. they upload the license
  a) using the ROR settings page
  b) using API (the one decided here)
 
So the facts are:
* we generate the license 
* user uploads it
* user doesn't have to know what implementation for the license structure

What could go wrong?
1. user copies part of the license (the common issue) - we should tell him: "we detect that sth is wrong with the license". There is no sense in telling them what, because they won't be able to fix it. At least we can advice him to check if he copied the whole string
2. user uses an expired license (a common issue too) - we should tell him: "hey, you used the expired license. But the new one for a god sake! :)". We don't need to tell them, that the JWT is expired. Their ROR license is expired. They context is an ROR license, not a JWT. 
3. user tries to modify the license (not so common thing) - we should tell him the same we do in the case of point 1. 

As you can see there is no need to bother them about the fact that JWT signature is bad, or JWT is invalid. 
This is why I think we should remove all the JWT-related things from this API.

.gitbook/assets/ReadonlyREST-api.yml
Dzuming: makes sense, changed ✔️

.gitbook/assets/ReadonlyREST-api.yml
Dzuming: Okay, so your proposition, is to remove states that are let's say, programmatic errors, because for some reason we generated a bad license key, and don't expose it to the end user, but still log it as a console error.  I misunderstood you here, I thought, that your proposed was about removing all 400 activation key based statuses. Changed ✔️

coutoPL: yes, it's still worth logging it for the support purposes 👍

coutoPL (approved):

PR #191, opened 2023-12-04
RORDEV-1025 clean stale sessions in case of index based sessions

Dzuming:
- Automatic Session cleanup description


sscarduzio:
session_timeout_minutes is quite an example because it's poorly named, and it's the poster child of the need for a overall restructuring of these settings. 

For now let's ship this because the customer is in trouble.

Dzuming:
> session_timeout_minutes is quite an example because it's poorly named, and it's the poster child of the need for a overall restructuring of these settings.

I added a task for it https://readonlyrest.atlassian.net/browse/RORDEV-1056 

> For now let's ship this because the customer is in trouble.
If you want to send a pre-release build there is a `1.54.0-pre1` version available

sscarduzio (approved):

PR #199, opened 2024-03-06
provide retrieval mode info

Dzuming:
provide retrieval mode info

### DIFF
--- a/universal-builds/universal-builds.md
+++ b/universal-builds/universal-builds.md
@@ -74,7 +74,7 @@ Setting this parameter effectively means that all other activation key retrieval
 
 #### Default behaviour
 
-If there is no `kibana.yml` license config provided or 
+If there is no `kibana.yml` license config provided or `activationKeyRetrievalMode: "all"` defined:
 
 ```yaml
 readonlyrest_kbn:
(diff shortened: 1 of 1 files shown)


universal-builds/universal-builds.md
```diff
@@ -66,3 +66,25 @@ This method is useful for manual deployments.
 1. Enter the activation key you obtained from the customer portal in the license management UI
 
 ![](assets/activation\_keys\_gui.png)
+
+### Change Activation key retrieval mode via kibana.yml
+
+In the kibana.yml configuration file, you have the option to specify the mode for retrieving activation keys. 
+Setting this parameter effectively means that all other activation key retrieval methods will be disregarded.
```
sscarduzio, line 73: Let's also state the default behaviour and the default order in which the retrieval methods are attempted ?

Dzuming: Done ✔️

universal-builds/universal-builds.md
```diff
@@ -72,15 +72,37 @@ This method is useful for manual deployments.
...
+
+If there is no `kibana.yml` license config provided or 
+
+```yaml
+readonlyrest_kbn:
+    license:
+        activationKeyRetrievalMode: "all" # "file" | "env" | "all" | "none"
+```
+
+defined, the order of AK checking the most recent AK is:
```
sscarduzio, line 85: "the most recent" ? Does it mean it checks in multiple places and elects the newest? Strange because later you say that we just stop evaluating when an AK is found?

Dzuming: Right, it's misleading. What I mean here, is that the order of activation key validation. I rephrased it

sscarduzio (approved):
Now I like it. Well done :)

PR #206, opened 2024-04-25
[RORDEV-1167][RORDEV-1121][RORDEV-1116] LDAP new features

coutoPL:
🧐Enhancement (ES) LDAP Connector feature: groups server-side filtering
🧐Enhancement (ES) LDAP Connector feature: skip user search option when user attribute is CN


elasticsearch.md
```diff
@@ -2131,6 +2132,7 @@ Usually, we would like to configure three main things for defining the way LDAP
       * `group_attribute_is_dn` (boolean, optional, default: `true`) -
         * when `true` the search filter will look like that: `(&YOUR_GROUP_SEARCH_FILTER(unique_member_attribute={USER_DN}))`
         * then `false` the search filer will look like that: `(&YOUR_GROUP_SEARCH_FILTER(unique_member_attribute={USER_ID_ATTRIBUTE_VALUE}))`
+      * `sever_side_groups_filtering` (boolean, optional, default: `false`) - by default ROR's LDAP connector asks for all groups of the given user. The group filtering is done on ROR's side. It allows ROR to cache them efficiently. But in some cases (e.g. when user has hundreds of groups), it's better to filter them on the LDAP server side. If this setting is `true`, LDAP will only be queried for a certain subset of the user groups (defined by the `groups_or`/`groups_and` subrule of the `ldap_authorization`/`ldap_auth` rule). Note, however, that ONLY the returned subset of the user's groups is cached.
```
sscarduzio, line 2135: This sounds like a great optimisation that I'd always want turned on. Can you make an example of use cases that would NOT benefit from having this feature turned on? Maybe a "in-depth" side-page with examples.

coutoPL: Yeah, it may sound like that but in our case, this is not always a good idea. I'll describe it in depth in a separate file, but not in this PR. I want to link this in the release changelog. So, I'll back to you when it's done

sscarduzio (approved):

PR #215, opened 2024-08-14
[RORDEV-617] ECK + docker images documentation

coutoPL:


examples/secure_metricbeats.md
```diff
@@ -0,0 +1,42 @@
+# Secure Metricbeats
```
Dzuming, line 1: Shouldn't it be Metricbeat?

image

kibana.md
```diff
@@ -1278,8 +1324,8 @@ Now try to login as user1, and ReadonlyREST Enterprise should initialize the ind
 
 You can configure the `number_of_shards` and `number_of_replicas` for the tenant index via the `kibana.yml` file, allowing you to override the default index settings. This can be particularly useful in a single-node environment.
 ```yaml
-readonlyrest.tenantIndex.number_of_shards: 1
-readonlyrest.tenantIndex.number_of_replicas: 0
+readonlyrest_kbn.tenantIndex.number_of_shards: 1
```
Dzuming, line 1327: This change was on the master branch, i merged it to the develop. Could you update your branch?

Dzuming (approved):
LGTM

sscarduzio:
> If running POD with root privileges is something you cannot accept, you can create your own image with the patching step done at the image creation level (not in the runtime as our image does) and place it your your registry

then why don't we provide this ourself?

> It's possible, but we don't recommend installing the ReadonlyREST plugin using the init-container pattern, because of the patching required patching step

This needs expansion. How? Why?

eck.md
```diff
@@ -0,0 +1,146 @@
+# ECK 
+
+ReadonlyREST plugins officially support installation on Elasticsearch and Kibana working in K8S cluster and managed by the [ECK operator](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-quickstart.html).
```
sscarduzio, line 3: Let's use "Kubernetes", the "k8s" is too informal.

eck.md
```diff
@@ -0,0 +1,146 @@
...
+        runAsNonRoot: false
+        runAsUser: 0
+        runAsGroup: 0
+      containers:
+        - name: kibana
+          env:
+            # we have to explicitly agree to patch the KBN binaries (the patching step will be done only once)
+            - name: I_UNDERSTAND_IMPLICATION_OF_KBN_PATCHING
+              value: "yes"
+            # we have to provide a ROR license if we want to use ROR Pro or Enterprise
```
sscarduzio, line 131: where do we get the keys? Link to the portal.
can we proceed without keys?

coutoPL:
> > If running POD with root privileges is something you cannot accept, you can create your own image with the patching step done at the image creation level (not in the runtime as our image does) and place it your your registry
> 
> then why don't we provide this ourself?
> 
> > It's possible, but we don't recommend installing the ReadonlyREST plugin using the init-container pattern, because of the patching required patching step
> 
> This needs expansion. How? Why?

Yeah, you are right. See how it looks now

coutoPL:
@sscarduzio WDYT?

eck.md
```diff
@@ -128,19 +137,176 @@ spec:
...
             - name: ROR_ACTIVATION_KEY
               value: "<YOUR_ACTIVATION_KEY/>"
 ```
 
 And these are all differences we need to make to run [ECK Quickstart](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-quickstart.html) with ReadonlyREST instead of X-Pack security. 
 
+## Using Docker images built by you and stored in your registry
+
+As you probably noticed, our docker images have to be run with root privileges. It's due to legal reasons and you, as a user, 
+have to confirm that you agree to do the patching steps. If running POD with root privileges is something you cannot accept, you can create your own image with the patching step done at the image creation level (not in the runtime as our image does) and place it your your registry.
```
sscarduzio, line 150: * running a pod
* not at runtime
* save it on your own registry

eck.md
```diff
@@ -128,19 +137,176 @@ spec:
...
+
+### Elasticsearch with ROR custom image
```
sscarduzio, line 152: this can be on its own page?

sscarduzio (commented):
Cool! Left some minor coments :)

eck.md
sscarduzio: The reason I wanted to isolate this special case into its own space is to keep the basic guide small and easy. People who want their own images bla bla it's something way more involved and I don't want people to get scared of it the first time they land here.

I guess the collapsible can do.

mk-software-pl: ok, done. Please, take a look

coutoPL: ok, done. please take a look

coutoPL: @sscarduzio is it ok now?

sscarduzio (approved):

PR #222, opened 2024-09-15
Update Client Authentication Methods

Dzuming:
- Move the Client Authentication Methods section from examples to Kibana plugin docs
- Provide the parameter name to the Client Authentication Methods description


coutoPL (approved):
LGTM

PR #232, opened 2024-11-07
RORDEV-1318 Patcher acknowledgement

Dzuming:
- Provide Patcher acknowledgement silent mode description


sscarduzio (commented):
"To apply patches in Kibana using a script in non-interactive mode (bypassing prompts), you have two options:"

Dzuming:
> "To apply patches in Kibana using a script in non-interactive mode (bypassing prompts), you have two options:"

Updated

sscarduzio (approved):

PR #237, opened 2024-11-14
add info about removed xpack parameters in Kibana 8.x

Dzuming:
- add info about removed xpack parameters in Kibana 8.x


examples/oidc-sso/keycloak_oidc.md
```diff
@@ -48,11 +48,11 @@ Then, configure the OpenID Connect (OIDC) client
 
 **kibana.yml** (without ssl enabled)
 
-```
+```yaml
 # More on how to enable SSL on the official documentation of Kibana
 server.ssl.enabled: false
 
-xpack.security.enabled: false
+xpack.security.enabled: false # Skip this setting in the Kibana configuration for version 8.x, as it has been removed.
```
coutoPL, line 55: WDYT about doing it the other way around?
The newest KBN version is a starting point. So, we should have this setting commented and the comment above it should say that it should be enabled for 7.x

Dzuming: Yes, in the places where we only have, the `kibana.yml` config it makes sense to do it this way, but in some areas, we have a config for both `kibana.yml` and `elasticsearch.yml` and IMO we cannot do it this way. So, maybe
1. For` kibana.yml` only places, let's comment it and change a comment to the `Uncomment this setting for a Kibana 7.x version`
2. For `kibana.yml` and `elasticsearch.yml` places, let's leave it as it is

WDYT?

coutoPL: I'm not sure if I understand why we cannot do that. Could you please show a specific example?

Dzuming: Sure,

here https://github.com/beshu-tech/readonlyrest-docs/pull/237/files#diff-c3178b08de0d2fc9f1fbbbf30a4c9fc9b0f0f13f01d083624fb186a00d327957L54
and here https://github.com/beshu-tech/readonlyrest-docs/pull/237/files#diff-e0655dbfd648d25cf494551f505f2e42c2e07f1f135db2313b1c87d2bf096ce3R746

coutoPL: oh, sure. It's ok. It will probably change when [this task](https://readonlyrest.atlassian.net/browse/RORDEV-1303) is done.

coutoPL (approved):

PR #246, opened 2025-02-21
[RORDEV-1388] Groups logic documentation

mgoworko:
## Summary by CodeRabbit

- **New Features**
  - Introduced advanced conditional rules for evaluating user group membership, including `groups_any_of`, `groups_all_of`, `groups_not_any_of`, `groups_not_all_of`, and a combined rule for flexible inclusion and exclusion criteria.
- **Documentation**
  - Updated access control guides with clear explanations and examples to support the enhanced group evaluation logic, reflecting the new `groups_any_of` structure across various sections.


elasticsearch.md
```diff
@@ -799,11 +799,86 @@ So that Kibana will forward the necessary headers to Elasticsearch.
 
 [Impersonation](details/impersonation.md) is supported by this rule without an extra configuration.
 
-#### `groups_or` (or `groups`)
+#### `user_belongs_to_groups`
```
coutoPL, line 802: Let's take a look at [this example](https://github.com/beshu-tech/ror-sandbox/blob/02859e22138bc5c0db7ef28a72d6cbed31fa249b/ror-demo-cluster/conf/es/readonlyrest.yml#L23). 

The "groups" rules use the users definition from the "users" section. In this section, we define [static users](https://github.com/beshu-tech/ror-sandbox/blob/02859e22138bc5c0db7ef28a72d6cbed31fa249b/ror-demo-cluster/conf/es/readonlyrest.yml#L30) (and we assign groups to them) or we can authorize [dynamic users](https://github.com/beshu-tech/ror-sandbox/blob/02859e22138bc5c0db7ef28a72d6cbed31fa249b/ror-demo-cluster/conf/es/readonlyrest.yml#L36) (and we can map the external groups to the local ones).

So, the first step of the "groups" rule is authorizing the user (after this step, we have an authorized user with assigned groups to which they belong). The second step is to check if the authorized user groups are permitted in context of the rule (and yes - the permission can be granted at the "users" section level too ... so, we can think about it as a second authorization).

So, `user_belongs_to_groups` in context of the above, is not a good choice. Maybe `permitted_groups` or `allowed_groups` would be better. @sscarduzio WDYT?

elasticsearch.md
```diff
@@ -799,11 +799,86 @@ So that Kibana will forward the necessary headers to Elasticsearch.
...
 
-`groups_or: ["group1", "group2"]`
+The ACL block will match when the user belongs to groups matching the condition defined in the `user_belongs_to_groups` config section. The information about what users belong to what groups is defined in the `users` section, typically situated after the ACL, further down in the YAML.
```
coutoPL, line 804: This should be changed too. See the comment above - seems that I wrote sth we can adapt and put here ;)

elasticsearch.md
```diff
@@ -799,11 +799,86 @@ So that Kibana will forward the necessary headers to Elasticsearch.
...
+##### `any_of`
+
+The ACL block will match when the user belongs to any of the specified groups.
+
+```yaml
+  user_belongs_to_groups:
+    any_of: ["group1", "group2"]
+```
+
+With deprecated syntax (without being wrapped inside the `user_belongs_to_groups` section):
```
coutoPL, line 815: In general, we don't mention the deprecated syntax. We do support it, but we want our users to migrate to the new one.

elasticsearch.md
```diff
@@ -799,11 +799,86 @@ So that Kibana will forward the necessary headers to Elasticsearch.
...
+- `groups_or: ["group1", "group2"]`
+- `groups: ["group1", "group2"]`
+
+##### `all_of`
+
+This rule is very similar to the above defined `any_of` rule, but this time ALL the groups listed in the array are required (boolean AND logic), as opposed to at least one (boolean OR logic) of the `any_of` rule.
```
coutoPL, line 821: subrule?

coutoPL, line 821: we can mention about the boolean OR logic in the any_of subrule too

elasticsearch.md
```diff
@@ -799,11 +799,86 @@ So that Kibana will forward the necessary headers to Elasticsearch.
...
+
+```yaml
+  user_belongs_to_groups:
+    all_of: ["group1", "group2"]
+```
+
+With deprecated syntax (without being wrapped inside the `user_belongs_to_groups` section):
```
coutoPL, line 828: the same as above

elasticsearch.md
```diff
@@ -799,11 +799,86 @@ So that Kibana will forward the necessary headers to Elasticsearch.
...
+##### `not_any_of`
+
+The ACL block will match when the user belongs to NONE of the specified groups.
+
+Looking at the example below:
+- ACL block will match for user that belongs to `group0`
+- ACL block will NOT MATCH for user that belongs only to `group1`
+- ACL block will NOT MATCH for user that belongs only to `group2`
+- ACL block will not match for user that belongs to both `group1` and `group2`
+```yaml
```
coutoPL, line 840: IMO the example should go first and then the explanations.

elasticsearch.md
```diff
@@ -799,11 +799,86 @@ So that Kibana will forward the necessary headers to Elasticsearch.
...
+
+##### `not_any_of`
+
+The ACL block will match when the user belongs to NONE of the specified groups.
+
+Looking at the example below:
+- ACL block will match for user that belongs to `group0`
+- ACL block will NOT MATCH for user that belongs only to `group1`
+- ACL block will NOT MATCH for user that belongs only to `group2`
+- ACL block will not match for user that belongs to both `group1` and `group2`
```
coutoPL, line 839: why sometimes "NOT MATCH" and sometimes "not match"

coutoPL, line 839: should we add the example with user with groups `group0` & `group1` & `group2`?

elasticsearch.md
```diff
@@ -799,11 +799,86 @@ So that Kibana will forward the necessary headers to Elasticsearch.
...
+```yaml
+  user_belongs_to_groups:
+    not_any_of: ["group1", "group2"]
+```
+
+With deprecated syntax (without being wrapped inside the `user_belongs_to_groups` section):
```
coutoPL, line 845: the same here

elasticsearch.md
```diff
@@ -799,11 +799,86 @@ So that Kibana will forward the necessary headers to Elasticsearch.
...
+
+##### `not_all_of`
+
+The ACL block will match when the user does not belong to all the specified groups.
+
+Looking at the example below (and notice the difference with `not_any_of` rule):
+- ACL block will match for user that belongs to `group0`
+- ACL block will MATCH for user that belongs only to `group1`
+- ACL block will MATCH for user that belongs only to `group2`
+- ACL block will not match for user that belongs to both `group1` and `group2`
```
coutoPL, line 856: the same comments as above:
1. example and explanations order
2. additional case

elasticsearch.md
```diff
@@ -799,11 +799,86 @@ So that Kibana will forward the necessary headers to Elasticsearch.
...
+```yaml
+  user_belongs_to_groups:
+    not_all_of: ["group1", "group2"]
+```
+
+With deprecated syntax (without being wrapped inside the `user_belongs_to_groups` section):
```
coutoPL, line 862: the same as above

elasticsearch.md
```diff
@@ -799,11 +799,86 @@ So that Kibana will forward the necessary headers to Elasticsearch.
...
+- `groups_not_all_of: ["group1", "group2"]`
+
+##### Combined rule
```
coutoPL, line 865: maybe `subrules combination`?

elasticsearch.md
```diff
@@ -799,11 +799,86 @@ So that Kibana will forward the necessary headers to Elasticsearch.
...
+
+The combined rule is a rule, that combines in a single ACL block both positive logic (`all_of`/`any_of`) with negative logic (`not_all_of`/`not_any_of`)
+The ACL block will match, when conditions of both rules are met.
+
+Looking at the example below:
+- ACL block will NOT MATCH for user that belongs only to `group0` (because the `any_of` logic is not satisfied)
+- ACL block will MATCH for user that belongs only to `group1` (the `any_of` logic is satisfied, the `not_all_of` too, because the user is not member of `group2`)
+- ACL block will MATCH for user that belongs to `group1` and `group3` for the same reason
+- ACL block will NOT MATCH for user that belongs to `group1` and `group2` (the `any_of` logic is satisfied, but `not_all_of` is not)
+```yaml
```
coutoPL, line 875: the same as above

elasticsearch.md
```diff
@@ -2611,7 +2686,8 @@ readonlyrest:
       group_names_claim: resource_access.client_app.group_names # optional, JSON-path style, see https://github.com/json-path/JsonPath
       header_name: Authorization
 ```
-You can verify groups assigned to the user with the `groups` field. The rule matches when the user belongs to at least one of the configured `groups` (OR logic). Alternatively, `groups_and` matches when the user belongs to all given groups (AND logic).
+
+You can verify groups assigned to the user with the groups logic (`any_of`/`all_of`/`not_any_of`/`not_all_of`/combined) described here [Groups logic](elasticsearch.md#user_belongs_to_groups)
```
coutoPL, line 2690: `described in the [Groups logic](...) section`

coutoPL (changes requested):
@mgoworko could you please confirm we use the new syntax in the whole docs?

elasticsearch.md
coutoPL: @sscarduzio kindly reminder. Is it ok for you, or maybe you have another idea?

coutoPL: oh, here we have to consider two cases:

1. the (local) groups rule:

```yaml
   - name: Accept requests from users in group team2 on index2
      groups_or: ["team2"]
      indices: ["index2"]
      
  users:
   ....
```

2. authorization rules (ldap_authorization, groups_provider_authorization, jwt_auth, ror_kbn_auth):
```yaml
ldap_auth:
  name: "ldap1"
  groups_or: ["group1", "group2"]
```

---

In this PR we introduced more advanced (and composable) groups logic. So we can do sth like this:

```yaml
   - name: Accept requests from users in group team2 on index2
      user_belongs_to_groups: 
        any_of: ["team2"]
        not_any_of: ["admins"]
      indices: ["index2"]
      
  users:
   ....
```

```yaml
ldap_auth:
  name: "ldap1"
  user_belongs_to_groups: 
        any_of: ["team2"]
        not_any_of: ["admins"]
```

---

Problems I see here:
1. I'm not sure if `user_belongs_to_groups` is the best choice. `groups` is too laconic. On the other hand, the proposed by me `allowed_groups` fits well in the context of local groups rule, but not in the second one
2. Maybe we have also a problem with the proper naming of the local groups rule. I always thought it's too little intuitive. It doesn't suggest we deal with the local groups and the simple RBAC in the "users" section.

WDYT guys?

elasticsearch.md
mgoworko: Added for both any_of and all_of

elasticsearch.md
mgoworko: I tried to adapt your description from the discussion above , please check

elasticsearch.md
mgoworko: I tried to come up with a name, that will be descriptive and will clearly inform, what are the conditions for the block to match. So I thought about it that way: 
```The block will match, when the user belongs to [any of] the groups [team1, team2]```

I am open to ideas about better name. As You wrote, `group` is not enough when we introduced more complicated logic/conditions. IMO the name `allowed_groups` when combined with negative logic, e.g. `not_any_of`  could potentially be more confusing than the current name.

Comment (coutoPL) [reply elasticsearch.md]: the decision related to this is [here](https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/pull/1080#discussion_r1986136646)

File: elasticsearch.md [hunk 14]
```diff
@@ -799,11 +799,108 @@ So that Kibana will forward the necessary headers to Elasticsearch.
 
 [Impersonation](details/impersonation.md) is supported by this rule without an extra configuration.
 
-#### `groups_or` (or `groups`)
+#### Groups subrules
```
coutoPL, line 802: Groups rules

elasticsearch.md
```diff
@@ -799,11 +799,108 @@ So that Kibana will forward the necessary headers to Elasticsearch.
...
 
-`groups_or: ["group1", "group2"]`
+The ACL block will match, when the user belongs to groups matching the specified conditions.
 
-The ACL block will match when the user belongs to any of the specified groups. The information about what users belong to what groups is defined in the `users` section, typically situated after the ACL, further down in the YAML.
+The groups subrules use the user definitions from the `users` section. In that section, we define static users (and we assign groups to them) or we can authorize dynamic users (and we can map the external groups to the local groups).
```
coutoPL, line 806: 1. subrules -> rules
2. let's add https://docs.readonlyrest.com/elasticsearch#users-and-groups link to "the `users` section"

elasticsearch.md
```diff
@@ -799,11 +799,108 @@ So that Kibana will forward the necessary headers to Elasticsearch.
...
+
+- the first step of the groups subrules is authorizing the user
+- after this step, we have an authorized user with information about the authorized groups to which the user belongs
+- then we check whether the authorized user groups are permitted in context of the rule
+
+##### `any_of`
```
coutoPL, line 812: `groups_any_of` (let's use the rule name here and below)

elasticsearch.md
```diff
@@ -799,11 +799,108 @@ So that Kibana will forward the necessary headers to Elasticsearch.
...
+
+The ACL block will match when the user belongs to any of the specified groups (boolean OR logic).
+
+Extended syntax:
+```yaml
+  user_belongs_to_groups:
```
coutoPL, line 818: `groups`

elasticsearch.md
```diff
@@ -799,11 +799,108 @@ So that Kibana will forward the necessary headers to Elasticsearch.
...
+    any_of: ["group1", "group2"]
+```
+
+Simplified syntax:
```
coutoPL, line 822: Let's show to simplified syntax first

elasticsearch.md
```diff
@@ -799,11 +799,108 @@ So that Kibana will forward the necessary headers to Elasticsearch.
...
+```yaml
+  groups_any_of: ["group1", "group2"]
+```
+
+##### `all_of`
+
+This subrule is very similar to the above defined `any_of` rule, but this time ALL the groups listed in the array are required (boolean AND logic), as opposed to at least one (boolean OR logic) of the `any_of` rule.
```
coutoPL, line 829: 1. subrule -> rule
2. any_of -> groups_any_of

elasticsearch.md
```diff
@@ -799,11 +799,108 @@ So that Kibana will forward the necessary headers to Elasticsearch.
...
+```
+
+Looking at the example above:
+- ACL block will MATCH for user that belongs to `group0`
+- ACL block will MATCH for user that belongs only to `group1`
+- ACL block will MATCH for user that belongs only to `group2`
+- ACL block will NOT MATCH for user that belongs to both `group1` and `group2`
+- ACL block will NOT MATCH for user that belongs to `group0`, `group1` and `group2`
+
+##### Subrules combination
```
coutoPL, line 886: `groups_combined`

elasticsearch.md
```diff
@@ -799,11 +799,108 @@ So that Kibana will forward the necessary headers to Elasticsearch.
...
+
+Subrules can be combined inside a single ACL block. It applies only to combining one positive logic subrule (`all_of`/`any_of`) with one negative logic subrule (`not_all_of`/`not_any_of`)
```
coutoPL, line 888: Let's rather talk about conditions that can be combined, rather than subrules.

elasticsearch.md
```diff
@@ -868,11 +959,17 @@ It handles LDAP authentication only using the configured LDAP connector (here `l
 ```yaml
 ldap_authorization:
   name: "ldap1"
-  groups: ["group3"]
+  user_belongs_to_groups:
```
coutoPL, line 962: 1. let's use simplified syntax 
2. I see we use deprecated `groups` syntax in our docs. Can we change it in all the docs to the new one?

elasticsearch.md
```diff
@@ -2611,7 +2708,8 @@ readonlyrest:
       group_names_claim: resource_access.client_app.group_names # optional, JSON-path style, see https://github.com/json-path/JsonPath
       header_name: Authorization
 ```
-You can verify groups assigned to the user with the `groups` field. The rule matches when the user belongs to at least one of the configured `groups` (OR logic). Alternatively, `groups_and` matches when the user belongs to all given groups (AND logic).
+
+You can verify groups assigned to the user with the groups logic (`any_of`/`all_of`/`not_any_of`/`not_all_of`/combined) described in the [Groups logic](elasticsearch.md#user_belongs_to_groups) section.
```
coutoPL, line 2712: use `groups_any_of` etc

coutoPL (changes requested):
and additional comments:
1. in some places of our docs we write about "groups rule". Maybe we should write about "Groups rules" instead (with the link to proper Groups rules section)
2. take a look at [this page](https://docs.readonlyrest.com/elasticsearch/groups-rule-mapping) too. IMO we need some adaptions also there

elasticsearch.md
mgoworko: Ad 2 - modified occurrences of groups_and/groups_or/groups all over the documentation - new changes that need to be reviewed.

coutoPL (changes requested):
1. let's change the request logs too. E.g.:
```text
FORBIDDEN by default req={ ID:747832602--1038482600#1312150, TYP:SearchRequest, CGR:N/A, USR:[no basic auth header], BRS:true, ACT:indices:data/read/search, OA:127.0.0.1, IDX:, MET:GET, PTH:/_search, CNT:<N/A>, HDR:Accept,content-length,Content-Type,Host,User-Agent,X-Forwarded-For, HIS:[::Infosec::->[groups->false]], [::KIBANA-SRV::->[auth_key->false]], [guest lol->[auth_key->false]], [::LOGSTASH::->[auth_key->false]], [::Infosec::->[groups->false]], [::ADMIN_GRP::->[groups->false]], [::Kafka::->[auth_key->false]], [::PERSONAL_GRP::->[groups->false]] }
```

2. IMO we should mention in each authorization rule that the groups logic syntax can be used there and (probably) link the "Checking groups logic" section.

mgoworko:
Ad1. I found only this one occurence and changed groups -> groups_any_of

Ad2. Added

PR #253, opened 2025-04-04
RORDEV-1262 Add documentation for extending Kibana API with x-ror-tenancy-id header

Dzuming:
## Summary by CodeRabbit

- **Documentation**
  - Introduced a new section in the documentation that explains how to extend the Kibana API using a custom HTTP header for tenant targeting.
  - Provided clear guidance and an example command for including the header in API requests to retrieve saved objects.

### DIFF
--- a/kibana.md
+++ b/kibana.md
@@ -545,7 +545,7 @@ You can configure an ACL in multi tenancy mode by adding a few ACL blocks contai
 ### Extending the Kibana API with the x-ror-tenancy-id header
 ([Enterprise](https://readonlyrest.com/enterprise))
 
-To target a specific tenant when making a [Kibana API](https://www.elastic.co/guide/en/kibana/current/api.html) request, include the custom HTTP header `x-ror-tenancy-id`. The value of this header should match one of the [groups rules](/elasticsearch#groups-rules) defined in your ACL configuration.
+To target a specific tenant when making a [Kibana API](https://www.elastic.co/guide/en/kibana/current/api.html) request, include the custom HTTP header `x-ror-tenancy-id`. The value of this header should match one of the [groups rules](/elasticsearch#groups-rules) id defined in your ACL configuration.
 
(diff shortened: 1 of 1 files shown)


kibana.md
```diff
@@ -542,6 +542,19 @@ readonlyrest_kbn.multiTenancyEnabled: false
 
 You can configure an ACL in multi tenancy mode by adding a few ACL blocks containing the `kibana.index` [rule](https://docs.readonlyrest.com/elasticsearch#kibana). See examples and further explanation under our [multi-tenancy guide](examples/multitenancy\_guide.md).
 
+### Extending the Kibana API with the x-ror-tenancy-id header
+([Enterprise](https://readonlyrest.com/enterprise))
+
+To target a specific tenant when making a [Kibana API](https://www.elastic.co/guide/en/kibana/current/api.html) request, include the custom HTTP header `x-ror-tenancy-id`. The value of this header should match one of the [groups rules](/elasticsearch#groups-rules) defined in your ACL configuration.
```
coutoPL, line 548: Let's refer to the group id. Because there is also a name of group

coutoPL, line 548: We should also describe what happens when no group id is passed. 

And what happens when given group id is forbidden (because in the metadata response is not such group)

coutoPL (changes requested):

kibana.md
coutoPL:
```
What is the correct answer? The first group defined in the ACL for a specific user?
```
ah, right - the ROR ES determines it. Yes, this is the correct answer ;)

```
Isn't it redundant info?
```
yes, you are right

Dzuming: > ah, right - the ROR ES determines it. Yes, this is the correct answer ;)

It's still worth mentioning here, right?

coutoPL: yeah, let's add this info.

PR #255, opened 2025-04-07
RORDEV-1262-all-tenants-endpoint

Dzuming:
- Provide all tenants endpoint to the Swagger
- provide artificial example endpoint with x-ror-tenancy-id

## Summary by CodeRabbit

- **New Features**
  - Updated the API specification to version 1.2.0.
  - Introduced a new endpoint to retrieve all tenants associated with the authorized user.
  - Added a demonstration endpoint supporting multi-tenancy header usage.
  - Enhanced API responses with detailed tenant information and access control examples.


.gitbook/assets/ReadonlyREST-api.yml
```diff
@@ -194,6 +194,55 @@ paths:
               examples:
                 UnsupportedMediaTypeResponseExample:
                   $ref: "#/components/examples/UnsupportedMediaTypeResponseExample"
+
+  /api/ror/settings/all_tenants:
```
coutoPL, line 198: WDYT about `GET /api/ror/user/tenants` instead?

.gitbook/assets/ReadonlyREST-api.yml
```diff
@@ -194,6 +194,55 @@ paths:
...
+          description: The request was forbidden by ReadonlyREST's ACL
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ForbiddenRorResponse'
+              examples:
+                GetAllTenantsViaApiEnterpriseOnly:
+                  $ref: "#/components/examples/GetAllTenantsViaApiEnterpriseOnly"
+
+  /api/example:
```
coutoPL, line 227: I'm not sure about it. TBH, I would be confused about what it demonstrates. 
The header can be passed to any Kibana REST API endpoints. So, maybe it's better to show some KBN endpoint here instead?

Dzuming: > So, maybe it's better to show some KBN endpoint here instead?

I had the same feeling about a single Kibana endpoint; it would also be confusing, why we described only this single endpoint, and is it a Kibana endpoint or not. I bet there is a high chance that, user will just ignore something like `api/index_management/indices`, because he will think that it's our ROR API endpoint, not something to demonstrate the header.

So maybe something like `api/*` will pay more attention?

.gitbook/assets/ReadonlyREST-api.yml
Dzuming: Yeah, sounds more rest and self-explanatory

.gitbook/assets/ReadonlyREST-api.yml
coutoPL: If we add some proper disclaimer in the description (explaining what we try to show), I think we can do it this way.

Dzuming: > If we add some proper disclaimer in the description (explaining what we try to show), I think we can do it this way.

Do you point to `api/*` here?

WDYT about this description?

```
summary: Example endpoint demonstrating multi-tenancy with the x-ror-tenancy-id header
description: >-
  This endpoint represents how all Kibana API requests can use the x-ror-tenancy-id header 
  for multi-tenancy support. The value should match one of your configured group IDs in the ACL. 
  If not specified, the first group defined in the ACL for the user is used as the default value.
```

.gitbook/assets/ReadonlyREST-api.yml
```diff
@@ -194,6 +194,55 @@ paths:
...
+    get:
+      summary: Example endpoint to demonstrate x-ror-tenancy-id header usage
+      description: This endpoint is a representation of the Kibana REST API endpoint to demonstrate the usage of the x-ror-tenancy-id header for multi-tenancy support.
+      tags:
+        - ReadonlyREST Enterprise license endpoints
```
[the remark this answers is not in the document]
Dzuming: Hmm, right, but on the other hand, we want to inform a user that this header is enterprise only. I will add this information to the description

.gitbook/assets/ReadonlyREST-api.yml
Dzuming: Done ✔️

.gitbook/assets/ReadonlyREST-api.yml
```diff
@@ -194,6 +194,54 @@ paths:
...
+        200:
+          description: Returns a list of tenants associated with the authorized user.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/GetAllTenants'
+              examples:
+                UpdateMainSettingsSuccessResponse:
+                  $ref: "#/components/examples/GetAllTenantsResponse"
+        403:
```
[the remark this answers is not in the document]
coutoPL: @Dzuming yeah, let's change it too. And in several places we say "all tenants". Let's change it too

coutoPL (approved):
one comment left. 
LGTM

.gitbook/assets/ReadonlyREST-api.yml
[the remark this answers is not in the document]
Dzuming: > @Dzuming yeah, let's change it too. And in several places we say "all tenants". Let's change it too

I'm not sure if I follow. What is a suggested value?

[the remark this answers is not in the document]
coutoPL: I was thinking about things like the following:
- `summary: Get all tenants for the authorized user`
- `All Tenants`
- `GetAllTenantsSuccessResponse`
- etc

As I said before: "all tenants" can be misleading. IMO, it's better to say "user's tenants"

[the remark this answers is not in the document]
Dzuming: > I was thinking about things like the following:
> 
> * `summary: Get all tenants for the authorized user`
> * `All Tenants`
> * `GetAllTenantsSuccessResponse`
> * etc
> 
> As I said before: "all tenants" can be misleading. IMO, it's better to say "user's tenants"

Sorry, I completely missed this comment https://github.com/sscarduzio/readonlyrest_kbn/pull/665#discussion_r2058972501, yeah, totally makes sense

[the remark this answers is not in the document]
Dzuming: Done ✔️

PR #261, opened 2025-04-28
RORDEV-1457 Fix documenaton broken links

Dzuming:


details/fls-engine.md
```diff
@@ -1,6 +1,6 @@
 # FLS engine
 
-Applicable in the context of the [`fields` rule](elasticsearch.md#fields)
```
coutoPL, line 3: is this intentional?

details/kibana-7.8.x-and-older.md
```diff
@@ -665,7 +665,7 @@ readonlyrest:
       signature_key: "my_shared_secret_kibana1_(min 256 chars)" # <- use environmental variables for better security!
 ```
 
-You may also use any custom claim from the OIDC `userinfo` token in ACL rules by using `{{jwt:assertion.<path_to_your_claim>}}` syntax. See the [dedicated section ](<../elasticsearch.md#Dynamic variables from JWT claims>)for more information. ( **TIP** : Do not forget the `assertion` prefix in front of you jsonpath. )
+You may also use any custom claim from the OIDC `userinfo` token in ACL rules by using `{{jwt:assertion.<path_to_your_claim>}}` syntax. See the [Dynamic variables from JWT claims section](elasticsearch##usage-examples) for more information. ( **TIP** : Do not forget the `assertion` prefix in front of you jsonpath. )
```
coutoPL, line 668: it the link correct?

details/fls-engine.md
Dzuming: Yes, `elasticsearch.md` will try to open github page with result 404 not found 

image

details/kibana-7.8.x-and-older.md
Dzuming: Yeah, the problem here is that this `Dynamic variables from JWT claims` anchor doesn't exist, because this element is on the fifth level header (which makes sense in this context), so I changed it to the closest anchor element, which is the usage-examples section

coutoPL (approved):
LGTM

PR #309, opened 2026-04-17
[RORDEV-1303]` xpack.security.enabled: false` is no longer required in ROR

coutoPL:
## Summary by CodeRabbit

* **Documentation**
  * Updated Elasticsearch version requirement to 6.7.0 or newer
  * Simplified installation instructions by removing X-Pack security disabling step
  * Reorganized changelog format for improved readability
  * Enhanced SSL and encryption configuration guidance with clearer setup options
  * Updated integration examples (OIDC, SAML, multitenancy) to reflect simplified setup process

### DIFF
--- a/details/kibana-7.8.x-and-older.md
+++ b/details/kibana-7.8.x-and-older.md
@@ -387,7 +387,6 @@ Open up `conf/kibana.yml` and add the following:
 xpack.graph.enabled: false
 xpack.ml.enabled: false
 xpack.monitoring.enabled: true
-xpack.security.enabled: false # this is fundamental!
 xpack.watcher.enabled: false
 
 # Kibana server use ::KIBANA-SRV:: credentials
--- a/elasticsearch.md
+++ b/elasticsearch.md
@@ -31,7 +31,7 @@ The following diagram models an instance of Elasticsearch with the ReadonlyREST 
 3. The HTTP stack in Elasticsearch parses the HTTP request
 4. The HTTP handler in Elasticsearch extracts the indices, action, request type, and creates a `SearchRequest` \(internal Elasticsearch format\).
 5. The SearchRequest goes through the ACL \(access control list\), external systems like LDAP can be asynchronously queried, and an exit result is eventually produced.
-6.7.0 The exit result is used by the audit event serializer, to write a record to index and/or Elasticsearch log file
+6. The exit result is used by the audit event serializer, to write a record to index and/or Elasticsearch log file
 7. If no ACL block was matched, or if a `type: forbid` block was matched, ReadonlyREST does not forward the search request to the search engine and creates an "unauthorized" HTTP response.
 8. In case the ACL matches a `type: allow` block, the request is forwarded to the search engine
 9. The Elasticsearch code creates a search response containing the results of the query
@@ -131,7 +131,7 @@ When prompted about additional permissions, answer **y**.
 
 #### 3. Patch Elasticsearch
 
-If you are using Elasticsearch 6.7.07.0 or newer, you need **an extra post-installation step**. Depending on the [Elasticsearch version](https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/blob/master/ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches), this command might tweak the main Elasticsearch installation files and/or copy some jars to `plugins/readonlyrest` directory.
+If you are using Elasticsearch 6.7.0 or newer, you need **an extra post-installation step**. Depending on the [Elasticsearch version](https://github.com/sscarduzio/elasticsearch-readonlyrest-plugin/blob/master/ror-tools-core/src/main/scala/tech/beshu/ror/tools/core/patches), this command might tweak the main Elasticsearch installation files and/or copy some jars to `plugins/readonlyrest` directory.
 
 ```bash
 jdk/bin/java -jar plugins/readonlyrest/ror-tools.jar patch --I_UNDERSTAND_AND_ACCEPT_ES_PATCHING=yes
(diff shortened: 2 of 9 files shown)


coutoPL:
@sscarduzio kindly reminder

sscarduzio:
## Review

This is a meaningful content update — the old "disable X-Pack Security" step is genuinely gone from ROR's requirements, and the new guidance (prefer XPack SSL when `xpack.security.enabled: true`, use ROR SSL only when it's `false`) matches what the plugin actually enforces today (`RorSslSettings.scala:94-95` rejects ROR SSL when XPack Security is enabled, and `:109` logs a deprecation when SSL is defined in `readonlyrest.yml`).

A few issues worth fixing:

### 🔴 Missing minimum ROR version

The new behavior — SSL loaded from `elasticsearch.yml` as the primary location, and ROR rejecting its own SSL when `xpack.security.enabled: true` — was introduced in **ROR 1.68.0** (commit `a29f58cef`, RORDEV-1520). Users on ≤1.67.x still need the old "disable xpack" step and the nested `readonlyrest.yml` style. The PR should either:
- Add a prominent note: applies to ROR ≥ 1.68.0; for earlier versions, SSL must be configured in `readonlyrest.yml` and `xpack.security.enabled: false` is still required, or
- State the minimum ROR version at the top of the Encryption section.

Without this, readers on older versions will hit the "Cannot use ROR SSL when XPack Security is enabled" error with no explanation.

### 🔴 Config location/style changed silently

The old examples used nested YAML inside `readonlyrest.yml`:

```yaml
readonlyrest:
  ssl:
    keystore_file: "keystore.jks"
```

The new examples use flat dotted keys inside `elasticsearch.yml`:

```yaml
readonlyrest.ssl.keystore_file: "keystore.jks"
```

This is a significant user-facing change. The `readonlyrest.yml` nested form is **deprecated but still supported** (see `RorSslSettings.scala:109` warning). Readers should be told (a) the old form still works with a deprecation warning, and (b) the SSL block is now supposed to move from `readonlyrest.yml` to `elasticsearch.yml` — this isn't obvious from the diff alone. Other docs still show the old form (e.g. `examples/secure_logstash.md:37-44`, `details/fips.md:35,42`, `examples/elastic-cloud-cluster-integration/details.md:141`) — either update those too or acknowledge both forms work.

### 🔴 Contradictory Kibana examples

The PR removed `xpack.security.enabled: false` from ES configs, but the OIDC/SAML Kibana examples still contain it:
- `examples/oidc-sso/keycloak_oidc.md:55`
- `examples/saml-sso/keycloak_saml.md:70`

Given the new narrative recommends keeping `xpack.security.enabled: true`, and ROR deactivates XPack's auth during patching regardless, these remaining lines in `kibana.yml` examples are inconsistent with the new message. Clarify whether Kibana still needs it disabled or remove/update those lines too.

### 🟡 Typos / small issues

- `elasticsearch.md:295`: `#### 6 Restart Elasticsearch.` — missing period after `6` (was `#### 6. Restart Elasticsearch.` in old file; edit dropped the dot).
- `examples/saml-sso/keycloak_saml.md:107`: "Configure SSL accoring to this guide." → `according`. Also "this guide" has no target — the old text was "so we modify the elasticsearch.yml" which at least pointed somewhere. Either link to the Encryption section or delete the dangling sentence.
- `elasticsearch.md:506`: "It works only for ES above 6.7.0" — inclusive or exclusive? The installation section says "ES 6.7.0 or newer" (inclusive). Make consistent.

### 🟡 "Why does this matter?" paragraph slightly overclaims

> "During its patching step, ReadonlyREST deactivates XPack Security's authentication and authorization features — these are replaced by ROR's ACL engine."

Worth a sanity check against what `ror-tools patch` actually does today — if it selectively disables only the HTTP filter chain (not, say, the internal `xpack.security.authc.realms` machinery used by internode transport), the sentence could mislead. If true as stated, ignore this comment.

### ✅ What looks good

- Core change (drop "5. Disable X-Pack security module" step) is correct — matches current ROR plugin behavior.
- The new decision table (`xpack.security.enabled: false` → ROR SSL, `true` → XPack SSL) accurately reflects `RorSslSettings.scala:93-98`.
- Linking to official Elasticsearch security-setup docs for the XPack SSL case is the right call.
- The removal of the outdated X-Pack disable step in `examples/elastic-cloud-cluster-integration/details.md`, `examples/multitenancy_guide.md`, and the OIDC/SAML ES sections is consistent.
- `6.5.x` → `6.7.0` version bump in the patch/unpatch sections is consistent with the tag `v1.68.0_es6.7.0` being the earliest for this behavior.

coutoPL:
> Missing minimum ROR version

Won't fix.

ROR documentation describes the newest ROR plugin only. It was always like that, and it's intentional - we maintain backward compatibility of our plugins' configurations because we want our users to be on the latest possible ROR version and upgrades easily. This approach enforces security. Because of that, we describe only the current state of the plugin (the last version). This also makes the documentation simpler and readable. The old state of documentation is available on GH, where users can refer to it by tags. 

> Config location/style changed silently

Nope, it doesn't.
Both are valid [YAML syntax and can be used interchangeably](https://www.elastic.co/docs/deploy-manage/deploy/self-managed/configure-elasticsearch#_config_file_format).

> Contradictory Kibana examples

It's a different setting. This is a Kibana setting, not an Elasticsearch one. 
But it looks like we can improve this part of the ROR documentation, too. Done

> Typos / small issues

Done

> "Why does this matter?" paragraph slightly overclaims
> If true as stated, ignore this comment.

It's true. Ignored

coutoPL:
@sscarduzio I cannot click "ask for rereview", so I'm kindly asking here :)

sscarduzio (approved):

PR #312, opened 2026-04-24
[RORDEV-1646] ROR startup settings documentation

coutoPL:
## Documentation
* Updated configuration documentation to clarify distinction between cluster-wide ACL settings and per-node settings.
* Added comprehensive section on ACL settings source configuration with selection and tuning guidance.
* Enhanced documentation with configurable polling intervals and startup retry behavior options.
* Refreshed configuration examples for request handling and response code settings.

### DIFF
--- a/elasticsearch.md
+++ b/elasticsearch.md
@@ -406,6 +406,16 @@ Creating a dedicated, lightweight ES node where to install ReadonlyREST:
 
 
 ## Elasticsearch Configuration
+
+ReadonlyREST uses two distinct kinds of settings:
+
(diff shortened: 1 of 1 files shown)


mgoworko (approved):
LGTM
(Should we also document the full scope of changes in the ES ROR - the full interchangeability of yaml/jvm opt/env as the source of settings?)

coutoPL:
> LGTM (Should we also document the full scope of changes in the ES ROR - the full interchangeability of yaml/jvm opt/env as the source of settings?)

IMO, no. 
The JVM options were used but were not documented. IMO, we could treat them as deprecated (we advised some of our users in the forum to use them, so we have to maintain them, but new users should use elasticsearch.yml to define them instead). The configuration using ENVs reflects the [ES feature](https://www.elastic.co/docs/deploy-manage/deploy/self-managed/install-elasticsearch-docker-configure), so we do support it, but there is no sense in mentioning it.

PR #320, opened 2026-06-09
[RORDEV-2016] Audit improvements

mgoworko:
## Summary by CodeRabbit

* **Documentation**
  * Updated the audit configuration guide with a new master switch, clarified ACL decision logging controls, and refined sink routing guidance.
  * Expanded block-level audit behavior documentation, including how audit disabling affects default ACL log emissions.
  * Reworked log output documentation with new logger routing guidance and added support for an ACL serializer (limited to log outputs).
  * Updated index and data stream output guidance, including delivery warnings and revised operational steps.
  * Refreshed audit serializer documentation, aligning allowed-event serialization behavior with block settings.

### DIFF
--- a/details/audit.md
+++ b/details/audit.md
@@ -37,70 +37,158 @@ Here is an example of the data points contained in each audit event. We can leve
 
 ## Configuration
 
-The audit collecting by default is disabled. To enable it, you need to add `audit.enabled: true` and optionally configure the `audit.outputs`.
-In the `outputs` array, you can define i.a. where the audit events should be sent.
-The currently supported output types are:
-* `index` - similarly to Logstash it writes audit events in the documents stored in the ReadonlyREST audit index
-* `data_stream` - similar to index type, but the audit events are stored in the ES data stream
-* `log` - it allows you to collect audit events using the Elasticsearch logs and format them with the help of features that `log4j2` enables.
+The audit outputs are disabled by default. To enable them, add `audit.enabled: true` and optionally configure `audit.outputs`.
 
-You can configure multiple outputs for audit events. When the audit is enabled, at least one output has to be defined.
-If you omit `outputs` definition, the default `index` output will be used.
+**Note**: Even when `audit.enabled` is `false` or not set, the built-in ACL log is a special case — it writes a human-readable decision line to Elasticsearch logs for every request by default. See [The default ACL log](#the-default-acl-log) below.
 
-Audit can also be controlled at the block level:
-- if audit is globally enabled, it is applied to all `access_control_rules` blocks by default — audit events will be generated for events pertaining to all blocks
-- the audit can be optionally disabled for individual blocks, as shown in the example below for the `Kibana` block
-- if audit is globally disabled, then it is disabled for all blocks, regardless of individual block settings
+When `audit.enabled: true` and no `outputs` are specified, ROR defaults to storing events in a local Elasticsearch index.
 
-**⚠️IMPORTANT**: When audit is disabled for a specific block, then there will be no audit events when that block is matched.
+### Global audit settings
 
-Here is an example of how to enable audit events collecting with all defaults:
 ```yaml
 readonlyrest:
-
   audit:
-    enabled: true 
+    enabled: true                   # enable/disable the entire audit subsystem
+    default_acl_log_enabled: true   # enable/disable the built-in ACL log (default: true)
+    outputs:
+    - type: index
+      name: my-index-sink           # optional name, used for per-block sink routing
+```
+
+| Setting | Default | Description |
+|---|---|---|
+| `enabled` | `false` | Master switch for the audit subsystem |
+| `default_acl_log_enabled` | `true` | Controls the built-in ACL log sink (see below) |
+| `outputs` | default index sink | List of audit output sinks |
+
+### The default ACL log
+
+When `default_acl_log_enabled: true` (the default), ROR writes a human-readable ACL decision line to Elasticsearch logs for every request, using the logger named `tech.beshu.ror.accesscontrol.logging.AccessControlListLoggingDecorator`. This happens regardless of whether any `outputs` are configured.
+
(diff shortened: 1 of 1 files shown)


details/audit.md
```diff
@@ -37,70 +37,158 @@ Here is an example of the data points contained in each audit event. We can leve
...
-In the `outputs` array, you can define i.a. where the audit events should be sent.
-The currently supported output types are:
-* `index` - similarly to Logstash it writes audit events in the documents stored in the ReadonlyREST audit index
-* `data_stream` - similar to index type, but the audit events are stored in the ES data stream
-* `log` - it allows you to collect audit events using the Elasticsearch logs and format them with the help of features that `log4j2` enables.
+The audit outputs are disabled by default. To enable them, add `audit.enabled: true` and optionally configure `audit.outputs`.
 
-You can configure multiple outputs for audit events. When the audit is enabled, at least one output has to be defined.
-If you omit `outputs` definition, the default `index` output will be used.
+**Note**: Even when `audit.enabled` is `false` or not set, the built-in ACL log is a special case — it writes a human-readable decision line to Elasticsearch logs for every request by default. See [The default ACL log](#the-default-acl-log) below.
```
coutoPL, line 42: Maybe when we talk about defaults, we could show what the explicitly defined equivalent would look like?

details/audit.md
```diff
@@ -37,70 +37,158 @@ Here is an example of the data points contained in each audit event. We can leve
...
+    default_acl_log_enabled: true   # enable/disable the built-in ACL log (default: true)
+    outputs:
+    - type: index
+      name: my-index-sink           # optional name, used for per-block sink routing
+```
+
+| Setting | Default | Description |
+|---|---|---|
+| `enabled` | `false` | Master switch for the audit subsystem |
+| `default_acl_log_enabled` | `true` | Controls the built-in ACL log sink (see below) |
```
coutoPL: We use "output" and "sink" words. Is this required or/and interntionl? Or maybe we can just use "output" in every place?

details/audit.md
```diff
@@ -37,70 +37,158 @@ Here is an example of the data points contained in each audit event. We can leve
...
+  - name: High-volume service account
+    auth_key: svc:secret
+    audit:
+      # send events only to the index, skip the ACL log line for this noisy block
+      enabled_audit_sinks: [my-index-sink]
+
+  - name: Admin users
+    auth_key: admin:admin
+    audit:
+      # keep both sinks active (default behaviour)
```
coutoPL: Do we miss sth here? I see "audit:" and the comment only

details/audit.md
```diff
@@ -37,70 +37,158 @@ Here is an example of the data points contained in each audit event. We can leve
...
   audit:
     enabled: true
-    outputs: [ index, data_stream, log ]
+    default_acl_log_enabled: false   # turn off the built-in ACL log
+    outputs:
+    - type: log
+      name: custom-acl-log
+      logger_name: my.custom.acl.logger
+      serializer:
+        type: acl                    # reproduces the built-in ACL log format
```
coutoPL, line 115: ok, but it would be nice to tell users about the format. Because I assume the format stays the same here.
And what bout the logger_name. What's this? Users may not know.

details/audit.md
```diff
@@ -37,70 +37,158 @@ Here is an example of the data points contained in each audit event. We can leve
...
+|---|---|---|
+| `enabled` | `true` | When `false`, no audit events are emitted when this block is matched, regardless of global settings |
+| `log_allowed_events` | `true` | When `false`, allowed requests matched by this block are not written to audit. Denied requests, errors, and index-not-found responses are always written |
+| `enabled_audit_sinks` | (all sinks) | Whitelist of sink names. Only the listed sinks receive events from this block. Use sink `name` values from `audit.outputs`, plus `default_acl_log` for the built-in ACL log |
+| `disabled_audit_sinks` | (none) | Blacklist of sink names. All sinks except the listed ones receive events from this block |
+
+`enabled_audit_sinks` and `disabled_audit_sinks` are mutually exclusive — you cannot specify both on the same block.
+
+**⚠️IMPORTANT**: When `audit.enabled: false` for a specific block, there will be no audit events at all when that block is matched — this suppresses both custom outputs and the default ACL log. **This is a change in behaviour from previous versions**, where block-level `audit: {enabled: false}` only suppressed the ES audit sinks while the ACL log continued to write.
```
coutoPL, line 146: But what about `verbosity`? I didn't check it yet, but I assume we have backward compatibility, so if someone declares `verbosity: info` explicitly, it should be treated as enabled ACL log with log_allowed_events: true` right?

details/audit.md
```diff
@@ -37,70 +37,158 @@ Here is an example of the data points contained in each audit event. We can leve
...
+
+```yaml
+readonlyrest:
   audit:
-    enabled: true 
+    enabled: true
+    default_acl_log_enabled: true
+    outputs:
+    - type: index
+      name: my-index-sink
```
coutoPL, line 88: Should we describe what this is? Is this required? If not, when users need it

details/audit.md
mgoworko: changed to `output` everywhere

details/audit.md
mgoworko: I think I meant an empty audit section, but that is not a very useful example, so now without `audit:` and with better comment.

details/audit.md
mgoworko: Added explanation what logger_name does and that acl serializer produces the same format as the built-in acl log

details/audit.md
mgoworko: added explanation

details/audit.md
mgoworko: 1. Yes, `verbosity: info` maps to `Block.Audit.Enabled(logAllowedEvents = true)` with all sinks active, so the ACL log continues to work exactly as before                              
2. The IMPORTANT note applies only when `audit: {enabled: false}` is set explicitly in a block, which maps to Block.Audit.Disabled and suppresses everything

Blocks using verbosity settings are unaffected - never mapped to Disabled.

coutoPL (approved):
LGTM

PR #328, opened 2026-08-02
[RORDEV-2134] Improve impersonation docs

mgoworko:
## Summary by CodeRabbit

* **Documentation**
  * Consolidated impersonation guidance into a comprehensive, self-contained guide.
  * Added configuration and Kibana workflows, support coverage, examples, troubleshooting, logging behavior, limitations, and glossary content.
  * Updated related authentication, authorization, Kibana, Test Settings, and external-service mock documentation to reference the guide.
  * Clarified hashed-credential limitations and added links to supported and unsupported rule sections.
  * Replaced duplicated page content with pointers to the centralized guide.


details/impersonation.md
```diff
@@ -22,6 +22,47 @@ Both of the problems can be solved using the ROR's impersonation. Thanks to the
 
 Admin can add the new user configuration without worrying and then test it by impersonating the user. They can check if the user can log in without problems and if the user has access only to the Kibana features the admin wanted to grant. When the admin is sure that everything is configured correctly, they can promote the settings (test) to production.
 
+## The impersonator/impersonated identity model
+
+Before diving into the configuration, it's worth understanding *what an impersonating request actually contains* and *what ROR does with it*, since this is the part of the configuration that is most error-prone.
+
+An impersonating request carries **two identities at once**:
+
+* the **impersonator** - the real, credentialed admin sitting behind the keyboard (e.g. `admin1`), proven by whatever credentials travel with the request (typically an HTTP Basic Auth header),
```
coutoPL: I'm not sure about `admin1` and `dev2` as impersonator and impersonatee:
1. Both are too technical, not natural
2. Impersonator doesn't have to be an admin (at least not in the ROR domain), so it may be misleading

Maybe we can use well-known `alice` and `bob`? WDYT?

details/impersonation.md
```diff
@@ -22,6 +22,47 @@ Both of the problems can be solved using the ROR's impersonation. Thanks to the
...
+* the **impersonated user** - the identity the admin wants to "borrow" for the duration of the request (e.g. `dev2`), carried in an internal header ROR/Kibana attaches to the request (`x-ror-impersonating`).
+
+ROR has to answer two completely different questions before it lets the request through:
+
+1. **"Is the real caller actually who they claim to be, and are they allowed to impersonate anyone at all?"** - this is a question about the *impersonator's* identity. It has nothing to do with `dev2`.
+2. **"Given that we trust the caller, are they allowed to become `dev2` specifically, and does `dev2` even exist?"** - this is a question about the *impersonated user*, answered using the `impersonation` section and, where needed, [service mocks](#defining-mocks-of-the-external-services-optional).
```
coutoPL: > , answered using the `impersonation` section and, where needed, [service mocks](#defining-mocks-of-the-external-services-optional).

I'm not sure if we need it in this place.

details/impersonation.md
```diff
@@ -22,6 +22,47 @@ Both of the problems can be solved using the ROR's impersonation. Thanks to the
...
+
+This is why the `impersonation` section needs its own, explicit `authentication_rule`, separate from whatever rule authenticates users in `access_control_rules`. **It's not accidental duplication - the two rules answer different questions, for different identities, and they run at different times:**
```
coutoPL: > This is why the `impersonation` section

We didn't mention the section yet. And we're answering the question that was never asked. It looks like this is not a proper place for this.

details/impersonation.md
```diff
@@ -22,6 +22,47 @@ Both of the problems can be solved using the ROR's impersonation. Thanks to the
...
+
+* The rule in `access_control_rules` authenticates whoever is trying to act as `dev2` during `dev2`'s own, non-impersonating session. During impersonation, ROR deliberately does **not** execute that rule's normal authentication logic - that's the entire point of impersonation: it lets an admin experience `dev2`'s permissions without needing `dev2`'s actual password, and without ROR having to make a call to `dev2`'s LDAP/external backend (that's also why [mocks](#defining-mocks-of-the-external-services-optional) exist - the impersonated identity's data is simulated, not fetched from the backend).
```
coutoPL: >  The rule in `access_control_rules` authenticate

The authentication rule ... which is a part of the block, which is one of many in the ACL.

coutoPL: The whole point is a bit complicated, I think.

IMO, we should describe it from a different point. We should make users aware that the impersonation request is almost the same as the request called by the real user. The only difference is the auth data it contains. So, the request sent by the impersonator on behalf of the user is processed by the ACL in the same way the request called by the user would be - with one difference: the way any authentication/authorization rule works. And the difference is crucial here.

details/impersonation.md
```diff
@@ -22,6 +22,47 @@ Both of the problems can be solved using the ROR's impersonation. Thanks to the
...
+* The `authentication_rule` inside the matching `impersonation` entry authenticates the **real caller** (`admin1`) using the credentials that are actually present on the wire. It has to be spelled out explicitly because there is no other rule anywhere in the ACL whose job is "verify this is really `admin1`, right now, for the purpose of impersonation" - the block rules are all written with the *impersonated* users in mind, not the impersonator.
+
+In practice, admins usually configure the same credentials/mechanism for `admin1` in both places, because `admin1` authenticates the same way whether they're doing their own work or impersonating someone. That similarity is exactly what makes the two entries *look* like copy-pasted duplication - but they are checked by different execution paths, at different moments, and there's nothing stopping you from requiring a different (e.g. stronger) authentication method just for impersonation.
```
coutoPL: We can imagine a case where we have an impersonator who cannot log in to Kibana. To impersonate using the ES REST API, we need to have the impersonator in the impersonators section only. But if we want him to be able to log into Kibana, it has to be authorized by ACL as a regular user. We should clearly say that if the request to ES is related to the `admin1` user, it has to be authorized by ACL. If `admin1` acts as an impersonator, it has to be authorized differently.

details/impersonation.md
```diff
@@ -22,6 +22,47 @@ Both of the problems can be solved using the ROR's impersonation. Thanks to the
...
+
+### What actually happens, step by step
+
+1. The HTTP request arrives carrying the impersonator's own Basic Auth credentials (`Authorization: Basic ...`) plus the `x-ror-impersonating: <target-username>` header.
+2. Before any ACL rule is evaluated, ROR decides which settings apply to the request. The mere presence of the `x-ror-impersonating` header makes ROR evaluate the request against **Test Settings** - never against Main Settings, even if Main Settings also happens to define an `impersonation` section of its own. If Test Settings aren't currently active (never applied, expired, or manually invalidated), the request is rejected immediately with `TEST_SETTINGS_NOT_CONFIGURED`, before any block gets a chance to run. See [Creating ROR's Test Settings](#creating-rors-test-settings) for how long Test Settings stay active.
```
coutoPL: > against **Test Settings** - never against Main Settings

We need a link here. I see we have one at the end of the section, but as a reader, at this point, I have a feeling that I don't understand what is the difference between Main and Test Settings. I want to know it now ;)

coutoPL: > TEST_SETTINGS_NOT_CONFIGURED

This is an internal detail. I'm not sure if we should mention it. It's not needed to understand what we want to describe.

details/impersonation.md
```diff
@@ -22,6 +22,47 @@ Both of the problems can be solved using the ROR's impersonation. Thanks to the
...
+3. ROR starts evaluating the Test Settings' `access_control_rules` blocks as usual, top to bottom.
+4. The moment ROR reaches an authentication rule that supports impersonation (`auth_key*`, `auth_key_unix`, `proxy_auth`, `token_authentication`, `ldap_authentication`, `external_authentication`, ...) inside a block, it notices the impersonation header and **does not run that rule's normal logic at all**. Instead, it switches into the impersonation flow below - which of those rules is written in the block doesn't change how the impersonator is identified. Rules that don't support impersonation (`jwt_*`, `ror_kbn_*` - see [Which rules support impersonation](#which-rules-support-impersonation)) never enter this flow: they evaluate the request exactly as they normally would, find no token in it, and their block simply doesn't match.
```
coutoPL: IMO, we should introduce the distinguishing point: not all auth rules support impersonation. And then explain:
1. what happens when it doesn't
2. what happens when it does

coutoPL: > ever enter this flow: they evaluate the request exactly as they normally would,

Are you sure? I thought they immediately fail and, as a result, the block in which they are fails to match. Not because the credentials are bad. Because we are in the impersonation flow, and the rule doesn't support it.

details/impersonation.md
```diff
@@ -22,6 +22,47 @@ Both of the problems can be solved using the ROR's impersonation. Thanks to the
...
+5. ROR extracts the impersonator's username from the request's Basic Auth header - this is always how the impersonator is identified, no matter which rule type the `impersonation` entry uses - and picks the **first entry whose `impersonator` pattern matches that username**. Only that one entry is then used: its `users` pattern is checked against the target username from the `x-ror-impersonating` header, and if it doesn't match, the request is denied with `IMPERSONATION_NOT_ALLOWED`. ROR does **not** continue scanning for a later entry that would allow the pair.
```
coutoPL: >  this is always how the impersonator is identified, no matter which rule type the impersonation entry uses - and picks the first entry whose impersonator pattern matches that username.

This is hard to understand.
Notice that we didn't introduce the `impersonation` section yet.

details/impersonation.md
```diff
@@ -22,6 +22,47 @@ Both of the problems can be solved using the ROR's impersonation. Thanks to the
...
+
+   This makes the order of the `impersonation` entries significant. If two entries have overlapping `impersonator` patterns (e.g. `admin*` and `admin1`), only the first matching one is ever consulted for a given impersonator - the `users` list of the later entry is dead configuration. Prefer one entry per impersonator, and put the most specific patterns first.
```
coutoPL: > Prefer one entry per impersonator, and put the most specific patterns first

I have a feeling that we mix two levels here:
1. high level of description
2. low level of configuration

We have to wisely arrange that:
1. we want to desibe the high level first
2. but we have to worry to not use concept that are not described yet - eg. in the low level
3. we have to worry to not mix these two things

details/impersonation.md
```diff
@@ -22,6 +22,47 @@ Both of the problems can be solved using the ROR's impersonation. Thanks to the
...
+
+   No entry matches the impersonator at all (or the request carries no Basic Auth header) → the same `IMPERSONATION_NOT_ALLOWED` denial, regardless of whether `admin1` is a perfectly valid, authenticated user elsewhere in the ACL.
```
coutoPL: > IMPERSONATION_NOT_ALLOWED

internal. Maybe we don't need to use it

details/impersonation.md
```diff
@@ -22,6 +22,47 @@ Both of the problems can be solved using the ROR's impersonation. Thanks to the
...
+6. ROR authenticates the request's Basic Auth credentials against **that entry's own `authentication_rule`** - a fresh, independent check, unrelated to the block ROR happened to be evaluating. Failure → `IMPERSONATION_NOT_ALLOWED`.
+7. ROR checks that the impersonator and the impersonated user aren't the same username (self-impersonation is rejected), and that the impersonated user actually exists. **The existence check is answered by the very rule ROR is currently evaluating**, using only what that rule knows: `auth_key: dev2:devpass` knows just `dev2`, an `ldap_authentication` rule asks the LDAP [service mock](#defining-mocks-of-the-external-services-optional), and so on. Three outcomes are possible:
```
coutoPL: > an `ldap_authentication` rule asks the LDAP [service mock]

e.g. we refer to the service mock. We didn't introduce it yet. Not even at the high level

coutoPL (changes requested):
I reviewed part of the changes.
I think we need to reorganise it a little bit, so I decided to stop. 

Please take a look at [this](https://docs.readonlyrest.com/examples/impersonation). 

The impersonation feature is rather the Kibana feature. It can be used only with ES (there are no restrictions to use it only with ES, but it would require to use our internal API, eg. the auth mocks API) but is not so user friendly and we don't describe it in our docs. 

Hisotrically, the feature was created in two steps - the ES support and the KBN support. The division was leaked to docs. IMO, we should merge it. We should describe the impersonation feature in one place (IMO [this one](https://docs.readonlyrest.com/examples/impersonation) is good) and the Kibana perspetive should be the first class citizen.

mgoworko:
> I reviewed part of the changes. I think we need to reorganise it a little bit, so I decided to stop.
> 
> Please take a look at [this](https://docs.readonlyrest.com/examples/impersonation).
> 
> The impersonation feature is rather the Kibana feature. It can be used only with ES (there are no restrictions to use it only with ES, but it would require to use our internal API, eg. the auth mocks API) but is not so user friendly and we don't describe it in our docs.
> 
> Hisotrically, the feature was created in two steps - the ES support and the KBN support. The division was leaked to docs. IMO, we should merge it. We should describe the impersonation feature in one place (IMO [this one](https://docs.readonlyrest.com/examples/impersonation) is good) and the Kibana perspetive should be the first class citizen.

I tried to go in that direction. I restructured the docs and applied the changes from other comments.

examples/impersonation/README.md
```diff
@@ -9,10 +9,315 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+1. Will the updated settings break anything on the production cluster?
+2. Is the new user configured correctly, with all the permissions they need and no more?
+
+Impersonation answers both. It always uses its own Test Settings, which are separate from the production settings, so you can change them without affecting other users. You add the new user to the Test Settings, impersonate them, and check that they can log in and see only the Kibana features you meant to give them. When everything is correct, you promote the Test Settings to production.
+
+## Impersonating a user in Kibana
+
+The workflow is available in the ROR menu, under **Edit security settings**. It has three steps:
+
+1. **Create Test Settings.** Impersonation doesn't use your production settings. It uses Test Settings: a separate, temporary ACL that you can edit freely and later promote to Main Settings. See [Creating Test Settings](test-settings-ui.md) for the UI, and [Creating ROR's Test Settings](#creating-rors-test-settings) for how Test Settings work.
```
coutoPL, line 37: > It uses Test Settings: a separate, temporary ACL that you can edit freely and later promote to Main Settings.

We should mention that the ACL related to the test settings won't be used by regular users. It is only used by the impersonators in the context of the impersonation session.

coutoPL, line 37: > See [Creating Test Settings](https://github.com/beshu-tech/readonlyrest-docs/blob/ea42fa7f5ae2625c5bbd031dbdffc624a29e6f8a/examples/impersonation/test-settings-ui.md) for the UI, and [Creating ROR's Test Settings](https://github.com/beshu-tech/readonlyrest-docs/blob/ea42fa7f5ae2625c5bbd031dbdffc624a29e6f8a/examples/impersonation/README.md#creating-rors-test-settings)

It looks like these two sections that describe test settings should be merged.

examples/impersonation/README.md
```diff
@@ -9,10 +9,315 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+2. **Define mocks of external services (optional).** This step is needed only if the users you want to impersonate come from LDAP or another external service. During impersonation, ROR doesn't connect to that service. Instead, it asks a mock which users exist and which groups they belong to, so you don't need a real account or password. See [Defining external services mock configurations](external-services-mocks-ui.md) and [Defining mocks of the external services](#defining-mocks-of-the-external-services-optional).
```
coutoPL, line 38: we should explain what "mock" means

coutoPL, line 38: The same about merging these two sections about external services mocks

examples/impersonation/README.md
```diff
@@ -9,10 +9,315 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+3. **Start impersonating.** Pick a user from the list, or type the username if it isn't listed. Kibana reloads as that user. See [Impersonating users](impersonate-user-ui.md).
+
+The list of users contains:
+
+* users defined statically in the settings,
+* users defined in the mocks of external services, such as LDAP.
```
coutoPL, line 44: Do we explain somewhere why we cannot get a list of, e.g., LDAP users? Or why we even need to define, e.g., an LDAP mock?

I'm not saying we have to do it here. But we have to explain it somewhere

examples/impersonation/README.md
```diff
@@ -9,10 +9,315 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+
+Users that come from upstream systems, for example through HTTP headers, aren't known in advance. To impersonate such a user, type their username.
```
coutoPL, line 46: An example?

examples/impersonation/README.md
```diff
@@ -9,10 +9,315 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+
+The workflow is available in the ROR menu, under **Edit security settings**. It has three steps:
+
+1. **Create Test Settings.** Impersonation doesn't use your production settings. It uses Test Settings: a separate, temporary ACL that you can edit freely and later promote to Main Settings. See [Creating Test Settings](test-settings-ui.md) for the UI, and [Creating ROR's Test Settings](#creating-rors-test-settings) for how Test Settings work.
+2. **Define mocks of external services (optional).** This step is needed only if the users you want to impersonate come from LDAP or another external service. During impersonation, ROR doesn't connect to that service. Instead, it asks a mock which users exist and which groups they belong to, so you don't need a real account or password. See [Defining external services mock configurations](external-services-mocks-ui.md) and [Defining mocks of the external services](#defining-mocks-of-the-external-services-optional).
+3. **Start impersonating.** Pick a user from the list, or type the username if it isn't listed. Kibana reloads as that user. See [Impersonating users](impersonate-user-ui.md).
+
+The list of users contains:
```
coutoPL, line 41: This shouldn't be a continuation of point 3?

examples/impersonation/README.md
```diff
@@ -9,10 +9,315 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+
+* users defined statically in the settings,
+* users defined in the mocks of external services, such as LDAP.
+
+Users that come from upstream systems, for example through HTTP headers, aren't known in advance. To impersonate such a user, type their username.
+
+The settings must also define who can impersonate whom. See [The impersonation section](#the-impersonation-section).
```
coutoPL, line 48: 1. what settings?
2. this is important information, but I'm not sure if it fits here

examples/impersonation/README.md
```diff
@@ -9,10 +9,315 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+
+## How ROR processes an impersonation request
```
coutoPL, line 50: We describe it from the perspective of the Kibana session.

In the previous point, we saw how to start the impersonation session. So, in this point, we can assume that the user (acting as the impersonator) is already logged in and is impersonating another user.

Let's start with this. Make the connection between this and the previous paragraph. And then write a little bit about internals - that, under the hood, Kibana sends requests to the ES cluster. And now you are ready to say that:

```yaml
Elasticsearch with the ReadonlyREST plugin handles an impersonation request almost the same way ...
```

examples/impersonation/README.md
```diff
@@ -9,10 +9,315 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+
+ROR handles an impersonation request almost the same way as a request sent by the impersonated user. It checks the request against the same ACL, block by block, in the same order. Rules such as `indices`, `kibana_*`, `fields`, `filter` or `hosts` work the same way as in that user's own session. This is what makes impersonation useful for testing settings.
```
coutoPL, line 52: > It checks the request against the same ACL, block by block,

maybe w link to the section in the elasticsearch.md where we describe it in detail (AFAIR, there is such a section somewhere)

coutoPL, line 52: >  testing settings.

testing ReadonlyREST ACL

examples/impersonation/README.md
```diff
@@ -9,10 +9,315 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+* The credentials in the request belong to the impersonator, for example `alice`. The name of the impersonated user, for example `bob`, is sent separately, in an internal header set by ROR and the ROR Kibana plugin.
+* Authentication and authorization rules behave differently.
+
+The sections below explain how.
+
+### Impersonating requests use Test Settings
+
+ROR has two separate sets of settings:
+
+* **Main Settings**: the ACL used for regular requests.
```
coutoPL, line 65: Previously, we talked about the production settings. Let's mention that here too

examples/impersonation/README.md
```diff
@@ -9,10 +9,315 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+* **Test Settings**: a separate ACL used only for impersonation. You can change Test Settings without affecting users who work with Main Settings, and promote them to Main Settings when you're done.
+
+ROR always checks impersonation requests against Test Settings, never against Main Settings. Test Settings are active only for a limited time and can be invalidated at any moment (see [Creating ROR's Test Settings](#creating-rors-test-settings)). When no Test Settings are active, impersonation doesn't work.
```
coutoPL, line 68: > impersonation doesn't work.

From the perspective of Kibana, impersonation is not available. Or impersonation cannot be used without them.

examples/impersonation/README.md
```diff
@@ -9,10 +9,315 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+### The impersonator and the impersonated user are authorized differently
+
+`alice` can appear in the ROR settings in two independent roles:
+
+* **As a regular user.** She is authenticated by the authentication rule of one of the ACL blocks, like any other user. This lets `alice` log into Kibana and do her own work.
+* **As an impersonator.** She is listed in the `impersonation` section, a separate part of the settings that defines who can impersonate whom and how each impersonator is authenticated (see [The impersonation section](#the-impersonation-section)).
+
+One role doesn't require the other:
+
+* An impersonator who never logs into Kibana needs only an entry in the `impersonation` section. They can still impersonate users through the Elasticsearch REST API.
```
coutoPL, line 79: We know it's possible, but we officially don't support it because we don't want our users to use our internal API directly.

examples/impersonation/README.md
```diff
@@ -9,10 +9,315 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+* If `alice` should also be able to log into Kibana as herself, an ACL block must authenticate her as a regular user. The `impersonation` section doesn't give her any access of her own.
+
+When `alice` sends a request as herself, the ACL handles it like any other request. When she sends a request as `bob`, the authentication rules in the ACL blocks no longer check who the caller is. The `impersonation` section does that instead. This is why each entry in the `impersonation` section needs its own authentication rule: it is the only place in the settings that verifies the impersonator.
```
coutoPL, line 82: Once again - the Kibana perspective. From the Kibana perspective, `alice` doesn't send any request

examples/impersonation/README.md
```diff
@@ -9,10 +9,315 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+
+### Not every authentication rule supports impersonation
+
+When an authentication rule receives an impersonation request, what happens next depends on whether the rule supports impersonation. The full list is in [Which rules support impersonation](#which-rules-support-impersonation).
```
coutoPL, line 86: rule doesn't receive, but rather processes a request

coutoPL, line 86: I wonder if we shouldn't list the auth rules that don't support impersonation instead of the list of supported ones. WDYT?

examples/impersonation/README.md
```diff
@@ -9,10 +9,315 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+
+**Rules that don't support impersonation** (`jwt_*`, `ror_kbn_*`) handle the request as usual. They look for a JWT or a ROR Kibana token, don't find one (the request contains the impersonator's Basic Auth credentials instead), and fail. The block doesn't match, and ROR moves on to the next block. Such blocks can't be tested with impersonation, and ROR shows a warning about them when Test Settings are applied.
```
coutoPL, line 88: > `jwt_*`, `ror_kbn_*`

We should not list them here. We should link the section mentioned above instead

coutoPL, line 88: > They look for a JWT or a ROR Kibana token, don't find one (the request contains the impersonator's Basic Auth credentials instead)

There are different reasons why some authentication rules don't support impersonation. We should clearly explain it in the section where we list the authn rules that don't support impersonation.

But this explanation (AFAIR) is not entirely true. When the request is an impersonation request and the rule that doesn't support impersonation processes it, the rule fails fast - it sees it's an impersonation request and it knows it doesn't support it, so it fails with IMPERSONATION NOT SUPPORTED internal cause (please verify).

examples/impersonation/README.md
```diff
@@ -9,10 +9,315 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+
+## Impersonation configuration
+
+This part describes the settings that the [Kibana workflow](#impersonating-a-user-in-kibana) depends on: Test Settings, the `impersonation` section, and mocks of external services such as [LDAP](../../elasticsearch.md#ldap-connector), [External Basic Auth](../../elasticsearch.md#external-basic-auth) or [Custom groups provider](../../elasticsearch.md#custom-groups-providers).
+
+### Creating ROR's Test Settings
+
+When you call Elasticsearch directly or through Kibana, ROR uses the ACL from Main Settings. Test Settings define a second ACL, which ROR for Elasticsearch uses only for requests that carry the impersonation header. ROR manages this header internally.
```
coutoPL, line 110: > that carry the impersonation header

The header is an internal detail. I'm not sure we should mention it. Let's find a better name to describe that request, e.g.,
* request with impersonation context
* request with impersonation data 
* etc

examples/impersonation/README.md
```diff
@@ -9,10 +9,315 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+
+**⚠️ IMPORTANT:** If an external service used in the ACL has no mock, ROR may report that impersonation is not supported. To avoid this, define mocks for all external services.
+
+The ROR Kibana plugin has a UI for creating and editing mocks. See the [service mock configuration guide](external-services-mocks-ui.md).
+
+### Which rules support impersonation
+
+Rules differ in how they support impersonation. When Test Settings are applied, ROR checks every rule and reports a warning for each block with a rule that won't work during impersonation. The ROR Kibana plugin shows these warnings in the Test Settings UI.
+
+| Rule | Impersonation support | Notes |
```
coutoPL, line 182: OK, I see the value in this table. Maybe we have to have two of them (the one for supporting and the one for not supporting impersonation)

examples/impersonation/README.md
```diff
@@ -9,10 +9,315 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+
+In Kibana, all logs of the impersonated user have this format: `[<log level>][plugins][ReadonlyREST][<filename>][impersonating <impersonated user username>]`
+
+When audit is enabled, the audit document contains an `impersonated_by` field.
+
+## Impersonation limitations
+
+Check whether these limitations affect your use cases:
+
+* Not everything in the ROR settings can be tested with impersonation, because some rules don't support it. For example, a rule with hashed credentials (such as `auth_key_sha512`) works during impersonation only in the `USER_NAME:HASH(PASSWORD)` form. If the username and password are hashed together, ROR can't read the username, and the rule won't match during impersonation. The [rules description](../../elasticsearch.md#rules) says which rules support impersonation.
```
coutoPL, line 288: The limitation is as follows: some rules simply don't support the implementation. We should link the table mentioned above, not repeat info from the table.

examples/impersonation/README.md
```diff
@@ -9,10 +9,315 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+* Test Settings are stored in the memory of the Elasticsearch node that received the save request from the ROR Kibana plugin, so impersonation works only on that node. For now, Kibana should communicate with only one Elasticsearch node. We plan to improve this in the future.
```
coutoPL, line 289: >  For now, Kibana should communicate with only one Elasticsearch node. We plan to improve this in the future.

Do you confirm that this is true? Don't we have automatic loading of the test settings on the other nodes? If yes, please create a Jira for this.

coutoPL (changes requested):

examples/impersonation/README.md
mgoworko: It's now the opening of "Defining mocks of the external services"

examples/impersonation/README.md
mgoworko: The list of users is now an indented paragraph under step 3, not a separate list after the workflow

examples/impersonation/README.md
mgoworko: Both fixed by moving it. The sentence is gone from the workflow, and its content is now part of "Impersonation configuration"

examples/impersonation/README.md
mgoworko: rewritten from that perspective

examples/impersonation/README.md
mgoworko: Done: "Main Settings: the production settings, with the ACL used for regular requests."

examples/impersonation/README.md
mgoworko: Changed to "impersonation is not available in Kibana"

examples/impersonation/README.md
mgoworko: Reframed. The paragraph now starts from the session ("During an impersonation session, the authentication rules in the ACL blocks no longer check who the caller is")

examples/impersonation/README.md
mgoworko: Now there are both, as two tables under "Which rules support impersonation":
- "Rules that support impersonation", with a "What it needs" column
- "Rules that don't support impersonation", with a "Why" and a "What you see" column

examples/impersonation/README.md
mgoworko: The page no longer mentions the header. It says "requests with impersonation context", and that ROR and the ROR Kibana plugin set that context internally.

examples/impersonation/README.md
mgoworko: That was my (and my agent's) mistake, we do support it, test settings are periodically reloaded

examples/impersonation/README.md
mgoworko: The jwt_* / ror_kbn_* rules does not have the information, that the request is an impersonating one. They look for a token, don't find one, and fail, so the block just doesn't match.

IMPERSONATION_NOT_SUPPORTED is the opposite case: a rule that does support impersonation in general, but can't finish the check in a specific case - a missing mock, hashed user:pass or impersonation disabled.

The new table gives the reason and the outcome for each one.

examples/impersonation/README.md
```diff
@@ -9,10 +9,356 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+* **The rule can't check.** This happens when a mock is missing, or when an `auth_key_sha*` rule hashes the whole `user:pass` pair, so the username can't be read from it. The block doesn't match, and ROR moves on to the next block. If no block matches `bob`, ROR reports that impersonation is not supported.
+
+A negative answer to any of the questions affects only the current block. ROR moves on to the next block, as it does for a regular request, and asks the questions again. The answers to the first two questions depend only on the `impersonation` section, so they are the same in every block. If one of them is negative, every block whose authentication rule supports impersonation fails, and the request is refused as not allowed, unless a block with no authentication rule matches it. The answer to the third question depends on the rule in each block, and it decides which block matches first.
+
+## Impersonation configuration
```
coutoPL: IMO, the things below could be a separate md file: e.g., "configuring the impersonation".

What's left in this file? The impersonation mechanism explanations without configuration details.

WDYT?

examples/impersonation/README.md
```diff
@@ -9,10 +9,356 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+
+![Auth mock](<../../.gitbook/assets/auth_mock.png>)
+
+After clicking the add/edit user buttons (1), you see a dialog where you can add (2) or remove (3) a user of an external service mock.
+
+![Add/edit external mock service](<../../.gitbook/assets/add_edit_auth_mock_service.png>)
+
+### Which rules support impersonation
+
+Rules differ in how they support impersonation. When Test Settings are applied, ROR checks every rule and reports a warning for each block with a rule that won't work during impersonation. The ROR Kibana plugin shows these warnings in the Test Settings UI.
```
coutoPL: We should state that all non-auth rules support impersonation requests.
But some of the auth rules don't.

examples/impersonation/README.md
```diff
@@ -9,10 +9,356 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+
+#### Rules that support impersonation
```
coutoPL: Auth rules

examples/impersonation/README.md
```diff
@@ -9,10 +9,356 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+          
+      ldaps:
+        - name: ldap1
+          [..]
+          
+      impersonation:
+        [...]
+    ```
+
+## Glossary
```
coutoPL: Maybe the glossary should stay in README.md

examples/impersonation/README.md
```diff
@@ -9,10 +9,356 @@ According to [Wikipedia](https://en.wikipedia.org/wiki/Impersonator):
...
+| A block that uses `jwt_auth` or `ror_kbn_auth` never matches during impersonation | No impersonation error, the block just doesn't match | These rules don't support impersonation. They look for a JWT or ROR Kibana token, don't find one, and fail. ROR shows a Test Settings warning for such blocks | Test these blocks in a real user session, or use a rule that supports impersonation |
+| The user you want to impersonate isn't on the list in Kibana | None (UI limitation) | The user matches only a wildcard pattern in the `users` section, so ROR can't list them | Type the username manually (see [limitations](#impersonation-limitations)) |
+| Impersonation is refused for every user, although the `impersonation` entry looks correct and the credentials work elsewhere | Impersonation not allowed | ROR identifies the impersonator only by HTTP Basic Auth credentials, whatever rule the `impersonation` entry uses. Credentials sent in any other form don't match any entry | Make sure the impersonator authenticates with a username and password (HTTP Basic Auth) |
+
+## Logs & audit
```
coutoPL: Maybe this section should stay in README.md. WDYT?

coutoPL (changes requested):

coutoPL:
and because we improve the docs, the target should be master

coutoPL (approved):
LGTM

PR #332, opened 2026-08-25
[RORDEV-2187] Use one word for an audit destination: output

coutoPL:
The audit documentation used two words for one thing. Some parts said "sink". Other parts said "output". A reader who sees two words looks for two things. There is only one thing.

The documentation now uses only "output".

## What changed

Only `details/audit.md`.

Block-level settings keys:

| Before | After |
|---|---|
| `enabled_audit_sinks` | `enabled_audit_outputs` |
| `disabled_audit_sinks` | `disabled_audit_outputs` |

The example output name `my-index-sink` is now `my-index-output`.

The text, the table and the heading "Per-sink routing example" now say "output".

## Why now

The code makes the same change. The two keys are on `develop`, but not on `master`. No release has them. No user must edit a configuration file.

### DIFF
--- a/details/audit.md
+++ b/details/audit.md
@@ -61,7 +61,7 @@ readonlyrest:
     default_acl_log_enabled: true   # enable/disable the built-in ACL log (default: true)
     outputs:
     - type: index
-      name: my-index-sink           # optional name, used for per-block sink routing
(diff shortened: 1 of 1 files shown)


mgoworko (approved):
LGTM

PR #333, opened 2026-09-20
[RORDEV-1861] Document failover mode and connectivity check for remote audit clusters

coutoPL:
The remote audit cluster got new settings: a failover mode and a connectivity check. The documentation did not describe them. It also said that ROR always sends the audit events in the round-robin mode. That is no longer true.

ES changes: sscarduzio/elasticsearch-readonlyrest-plugin#1222

## What changed

Only `details/audit.md`, both "Custom audit cluster" sections.

The `cluster` setting now has two syntaxes:

```yaml
# simple - round-robin, no connectivity check
cluster: ["https://user:password@auditNode1:9200"]

# extended
cluster:
  nodes: ["https://auditNode1:9200", "https://auditNode2:9200"]
  mode: failover
  username: "audit_user"
  password: "audit_password"
  connectivity_check: required
```

New settings:

| Setting | Required | Default | Values |
|---|---|---|---|
| `mode` | yes | - | `round-robin`, `failover` |
| `username`, `password` | no | credentials from the node URIs | - |
| `connectivity_check` | no | `disabled` | `required`, `best_effort`, `disabled` |

The page now also describes:

- How the two cluster modes pick a node, and when to use which one.
- Which failures move the request to the next node in the failover mode.
- Where the credentials go, and that all the nodes must use the same ones.
- What the connectivity check does at the settings loading: the retries, the timeout, the warning for a partly reachable cluster, and the rejection of nodes from different ES clusters.
- That the `data_stream` output always needs a reachable audit cluster, because it creates the data stream upfront.

## Summary by CodeRabbit

- **Documentation**
  - Expanded custom audit cluster documentation to cover simple and extended `index` output syntax.
  - Documented node selection options, shared credentials, and connectivity-check settings, including retries, timeouts, status codes, and cluster UUID handling.
  - Updated data-stream output documentation to use the same cluster syntax.
  - Clarified required connectivity and data-stream verification behavior.


mateuszkp96 (approved):
LGTM

